Issue #9, opened 2015-11-19
Grammar needs to support 'async'/'await' keywords


zhengbli:
Ported from https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/374

![image](https://cloud.githubusercontent.com/assets/972891/10553851/f5a45ad4-7412-11e5-9fdd-af3ed7a596ed.png)

DanielRosenwasser:
Closed by #20 thanks to @ejsmith.

PR #18, opened 2015-12-02
Add support for async / await syntax highlighting

ejsmith:
Not sure if this is complete, but it seems to work pretty well for us.

### DIFF
--- a/TypeScript.tmLanguage
+++ b/TypeScript.tmLanguage
@@ -116,7 +116,7 @@
 		<key>cast</key>
 		<dict>
 			<key>begin</key>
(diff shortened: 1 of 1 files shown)


ejsmith:
Fixes #9

zhengbli:
Thanks for the contribution! It seems that CI test has some issues, I'll fix soon.

DanielRosenwasser:
Since this only updates the XML files, you'll need to modify the YAML files which are the actual inputs.

ejsmith:
Doh! I will get that updated. What tool is used to generate the xml file?

DanielRosenwasser:
I might be wrong, but last I worked on these files, it was [AAAPackageDev](https://packagecontrol.io/packages/AAAPackageDev), a Sublime Text Plugin.

ejsmith:
Adding a new pull request with the yaml updated.

PR #33, opened 2016-01-05
add 'as' keyword

Gheoan:


DanielRosenwasser:
One issue with this is that we should be classifying the RHS as a type. @zhengbli can weigh in on whether that's a big deal.

zhengbli:
@Gheoan Because the things after `as` should be considered as types, can you try to add more rules for it? Something like:
```yaml
  as-cast:
    name: as-cast.expr.ts
    begin: \s*\b(as)\s+
    beginCaptures:
      '1': { name: keyword.operator.ts }
    end: '(?=[(;]|$)'
    patterns:
    - include: '#type'
```
and then add the `as-cast` to the `expression` rule set:
```yaml
  expression:
    name: meta.expression.ts
    patterns:
    ...
    - include: '#cast'
    - include: '#as-cast'  # newly added line
    - include: '#new-expr'
    ...
```
Otherwise it might not work in more complex cases. For example:
```ts
var s2 = t as (x: number) => string
var s3 = t as { foo: string, bar: number | symbol }
```
wouldn't get colored right without the rules for the part after the `as` keyword.

Gheoan:
This PR is for the `as` keyword used in [Import Declarations](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#1132-import-declarations). Does the `as` keyword have other uses in TypeScript that I am not aware of ?

zhengbli:
The `as` keyword can also be used for type casting. e.g.
```typescript
interface Foo {
	bar?: string
} 

var t = {} as Foo;
```

Gheoan:
@zhengbli done

DanielRosenwasser:
Hey @Gheoan can you update your PR to resolve merge conflicts?

Gheoan:
@DanielRosenwasser Done.

zhengbli:
@Gheoan Thanks for your contribution! And actually I just noticed you didn't sign the cla. For the formality, could you sign it for this and the other PR of yours? Thank you

Gheoan:
@zhengbli I had the impression that I signed it already. I signed it now properly.

PR #34, opened 2016-01-05
add 'from' keyword

Gheoan:


TypeScript.tmLanguage
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
```
zhengbli: Why is this change?

Gheoan: No idea. PackageDev generates it like that. Maybe because I'm on Windows. I will manually remove it.

Gheoan:
@zhengbli rebased and rady to merge

zhengbli:
@Gheoan Thanks!

Issue #53, opened 2016-01-24
Test code is broken (and 3 cases actually failing)


stkb:
Currently the test code is checking none of the test cases except the last one, through a classic closure mistake in [tests/test.ts](https://github.com/Microsoft/TypeScript-TmLanguage/blob/d563cc1e1737a5927608f251c952d748861672df/tests/test.ts#L13-L23):

``` typescript
describe("Compare generated with baselines", () => {
    for (var file of fs.readdirSync('generated')) {
        var generatedText = fs.readFileSync('generated/' + file, 'utf8');
        var baselinesText = fs.readFileSync('baselines/' + file, 'utf8');
        describe('Comparing ' + file, () => {
            it('should be the same', () => {
                chai.expect(compareText(generatedText, baselinesText)).to.true;
            })
        })
    }
})
```

Only the last generated values of ```generatedText``` and ```baselinesText``` will be checked against other each time. In actual fact these tests are failing:
* FunctionMethodOverloads
* Issue156
* Issue335

It's is easy enough to fix (the test code, not the failing tests) and I'm happy to submit a PR too but I wanted to add some other issues I came across while trying to make and test some changes to the syntax files:
* The tests won't run without first manually downloading the *.d.ts files with ```tsd install```. This should be automated (probably as part of ```npm install```), or documented.
* The ```/tests/typings``` folder should be added to ```.gitignore```
* Again with the typings/dependencies, ```assertion-error.d.ts``` doesn't seem to be used and the ```chai``` module is barely used. Not if there were other things planned for these.

stkb:
BTW, the issue can be fixed with something like this:
``` typescript
describe("Compare generated with baselines", () => {
    for (var file of fs.readdirSync('generated')) {
        var generatedText = fs.readFileSync('generated/' + file, 'utf8');
        var baselinesText = fs.readFileSync('baselines/' + file, 'utf8');

        addTestCase(file, generatedText, baselinesText);
    }
})

function addTestCase(file: string, generatedText: string, baselinesText: string) {
    describe('Comparing ' + file, () => {
        it('should be the same', () => {
            chai.expect(compareText(generatedText, baselinesText)).to.true;
        })
    })    
}
```

DanielRosenwasser:
Hey @stkb, thanks for the heads up! While @zhengbli is much more familiar with the infrastructure in place here, I think we'd be very willing to take a PR if you're up to it.

stkb:
[...]

PR #54, opened 2016-01-25
Fix testing code to check all tests, not just the last

stkb:
Fixes the testing code; previously it was only checking the last pair of files (see #53).

Note: the AppVeyor build for this PR should **fail** because now 3 tests are actually failing.


zhengbli:
:+1: I will send another PR to address the failed tests. Though it is important to fix the infrastructure first. Thank you @stkb !

stkb:
@zhengbli Thank you too :)

PR #55, opened 2016-01-25
Misc improvements to testing infrastructure

stkb:
Makes a few changes to try to reduce a bit of friction when it comes to running tests, for newcomers to the repository (as mentioned in #53)
* .gitignore
  * Adds ```tests/typings``` folder, which is generated automatically.
* tests/package.json
  * Adds ```tsd``` as a dev dependency and installs the *.d.ts files automatically as a post-install script. (BTW, you don't need to add "./node_modules/.bin/" inside npm scripts, as [npm does this anyway](http://firstdoit.com/npm-scripts/))
  * Adds the ```private``` flag (since the package will never be published and it gets rid of some npm warnings)
* README.md
  * Adds a brief few lines explaining how to run the tests.
* appveyor.yml
  * I added this because the build & test script would now be slightly simpler. I think it's better to have this in the open instead of in the settings of the AppVeyor project, but the current setup will still work, so it's up to you.


zhengbli:
:+1: @stkb Thanks for the contribution!

PR #48, opened 2016-01-15
Match EOL with various selectors

blakeembrey:
Not everyone uses semicolons, but right now some of the selectors assume usage of semicolons to terminate.

**After:**

image

image

**Before:**

image

image

PR #56, opened 2016-01-27
Fix failed tests

zhengbli:
Due to the previous error in test script, we have 3 tests are actually failed currently. This PR fixes these failed tests and added a new test for PR #48 which was merged when the CI wasn't working properly.


zhengbli:
@DanielRosenwasser @stkb comments?

TypeScript.YAML-tmLanguage
```diff
@@ -410,6 +402,39 @@ repository:
     - include: '#switch-case'
     - include: '#control-statement'
 
+  switch-variable:
```
DanielRosenwasser: switch-expression

TypeScript.YAML-tmLanguage
```diff
@@ -410,6 +402,39 @@ repository:
...
+    - include: '#expression'
+
+  switch-block:
+    name: switch-block.expr.ts
+    begin: '{'
+    end: (?=\})
+    patterns:
+    - include: '#expression'
+
+  switch-case:
```
DanielRosenwasser: case-clause

TypeScript.YAML-tmLanguage
```diff
@@ -410,6 +402,39 @@ repository:
...
+    name: case.expr.ts
+    begin: '(?<!\.)\b(case|default)\b'
+    beginCaptures:
+      '1': { name: keyword.control.ts }
+    end: ':'
+    patterns:
+    - include: '#expression'
+
+  switch-expression:
```
DanielRosenwasser: switch-statement

TypeScript.YAML-tmLanguage
```diff
@@ -169,7 +160,7 @@ repository:
       '5': { name: keyword.operator.ts }
       '6': { name: entity.name.function.ts }
       '7': { name: keyword.operator.ts }
-    end: '(?=\}|;|,|$)|(?<=\})'
```
DanielRosenwasser: Hmm, the thing is, a method declaration might not have a body and might just need on a newline. For instance,

```ts
abstract class C {
    abstract foo()
    abstract bar()
}
```

or with a method signature:

```ts
interface I {
    foo(): number
    bar(): string;
}

Comment (DanielRosenwasser) [inline TypeScript.YAML-tmLanguage]: Why was this changed? For braces on the next line?

Comment (zhengbli) [inline TypeScript.YAML-tmLanguage]: This change because the current definition breaks the following case:
```ts
class Test {
	acid(trip: string | {}, buttweasel: string) 
	{ // here
	}
}
```
because it thought the method definition ends at new line

DanielRosenwasser: Hm - I guess there's not really a problem, since you'll just highlight as a block anyway. This should be fine

zhengbli: Not it would not be fine. The block could be reorganized as an object body, and it would be hard to differentiate these two cases:
```ts
abstract class C {
    abstract foo() :
    {// object return type
    }
}
```
and 
```ts
abstract class C {
    abstract foo()
    {// method implementation
    }
}
```

zhengbli:
@DanielRosenwasser any further comments? Added the two cases you mentioned to the test.

DanielRosenwasser:
:+1:

TypeScript.YAML-tmLanguage
DanielRosenwasser: Sorry, by fine I meant your changes should be fine, not the code as it was originally written.

PR #57, opened 2016-01-27
Fix issue #16

### LINKED ISSUE #16: Variable declaration list: only 1st variable colored
Problem found using vscode v0.10.2
Variables after the first in a list of declarations don't get the special color applied.

declarations

Note: Not all themes support this extra color.

zhengbli:
Fix #16

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -11,24 +11,25 @@ patterns:
 repository:
   var-expr:
     name: meta.var.expr.ts
-    begin: (?<!\()\s*\b(var|let|const(?!\s+enum))
+    begin: (?<!\()\s*\b(var|let|const(?!\s+enum)\b)
     beginCaptures:
       '1': { name: storage.type.ts }
-    end: '(?=$|[;=\}\{])|(?<=\})'
+    end: (?=$|;)
     patterns:
-    - include: '#var-variable'
+    - include: '#var-single-variable'
     - include: '#comment'
 
-  var-variable:
-    name: meta.var-variable.expr.ts
-    begin: \b([a-zA-Z_$][\w$]*)
+  var-single-variable:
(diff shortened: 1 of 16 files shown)


stkb:
I didn't see any tests that covered these but from trying it myself, it looks like it's not completely fixed.  Without assignments it works: 

```
var ^a, ^b;
```

```
[1, 5]: source.ts meta.var.expr.ts meta.var-variable.expr.ts variable.ts 
[1, 8]: source.ts meta.var.expr.ts meta.var-variable.expr.ts variable.ts 
```

But add an assignment and it doesn't:
```
var ^a = 1, ^b;
```

```
[1, 5]: source.ts meta.var.expr.ts meta.var-variable.expr.ts variable.ts 
[1, 12]: source.ts 
```

zhengbli:
@stkb Thanks for the code review. I added test `multipleVariableDeclaration.ts` and updated the baselines. Any other comments?

stkb:
@zhengbli This is great and you certainly added a lot of test cases, which is why I'm sorry to nitpick but from trying it on my own code I found a couple of times where it still doesn't work:

If you split declarations over multiple lines:
```
var ^a,
  ^b;
```

Some people also use this style:
```
var ^a = 1
  , ^b = 2
```

zhengbli:
Yes it doesn't have good support for lines not ended with semicolons, which has always been difficult to get right. We tried before to not treat `$` as the ending sign, which caused a lot of troubles.

stkb:
I agree, I did look into tackling this myself but javascript's automatic semicolon insertion makes it hard :confused:

zhengbli:
@stkb I would say if there are no other issues other than this, I would merge in this change, and later if people complain about the problem I can tackle it separately.

stkb:
@zhengbli Yeah I'm fine with that :thumbsup:

PR #75, opened 2016-02-04
Addresses await keyword coloring.

### LINKED ISSUE #74: `await` colored incorrectly when part of a larger word.
Currently, `await` is colored as a keyword when inside of a larger word, such as `awaiting`.

rbuckton:
Fixes #74.


rbuckton:
@zhengbli, @billti: What do you think of this change?

DanielRosenwasser:
:+1:

PR #76, opened 2016-02-04
Changes scope for namespace/module to storage.type

rbuckton:
Changes scope for namespace/module to storage.type to better reflect its semantic meaning


rbuckton:
@zhengbli, @billti: What do you think of this change?

billti:
Seems like a reasonable and safe change. I assume you've played around with it and it looks good (both aesthetically and functionally)? If so, :+1:

Issue #90, opened 2016-02-19
Carets aren't colorized as operators


DanielRosenwasser:
The bitwise XOR operator (`^`) isn't classified as an operator like `&` or `|`.

PR #94, opened 2016-02-22
Fix for adding caret as a Logical Operator

anubmat:
FIXED ISSUE #90


DanielRosenwasser:
@anubmat this should have some test coverage associated with it too. This might be a problem since the tests use carets as markers.

DanielRosenwasser:
:+1:

Issue #63, opened 2016-01-28
Generic Types


blakeembrey:
Syntax highlighting on generic `type` keyword appears to be falling down.

![image](https://cloud.githubusercontent.com/assets/1088987/12660619/6ee4766a-c5c9-11e5-98ee-19af7701650f.png)

Without generic:

![image](https://cloud.githubusercontent.com/assets/1088987/12660624/78c989fe-c5c9-11e5-9ca0-95e10645df00.png)

PR #95, opened 2016-02-23
Fix for issue with type parameters on type alias

anubmat:
Fixing Issue #63


TypeScriptReact.YAML-tmLanguage
```diff
@@ -46,13 +46,16 @@ repository:
      - include: '#enum-declaration'
 
   type-declaration:
-    name: meta.type.declaration.tsx
-    begin: \b(type)\b\s+([a-zA-Z_$][\w$]*)\s*=\s*
+    name: meta.type.declaration.ts
```
zhengbli: All grammar defined in TypeScriptReact need to have a top level scope of `.tsx`
This change is unneeded.

zhengbli:
Please add tests to the PR. In the tests consider the following cases:
1. `<` and `>` used as "larger than" and "smaller than" signs
2. `<` and `>` used as opening and closing brackets for generic types

There was a discussion about these two cases before [at the Sublime repo](https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/285). And Bill has provided some test cases in his [comment](https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/285#issuecomment-126515395).

tests/cases/Issue63.ts
```diff
@@ -0,0 +1,7 @@
+^^type ^^Callback <^^string> = ^^(error: Error, value: T) => any
```
DanielRosenwasser: `string` isn't a valid type parameter name, so can you adjust the test?

DanielRosenwasser:
Can you add a test with type literal syntax like so?

```ts
^type ^^Foo<^^T> = ^{
    ^foo: ^T,
    ^bar: ^number
^}
```

zhengbli:
:+1:

PR #99, opened 2016-02-24
Correctly classify keywords in heritage clauses

### LINKED ISSUE #89: Typescript highlighting after "extends" keyword
_From @mattacosta on February 16, 2016 0:16_

It seems everything after "extends" turns into an identifier, when it should just be the very next word since there's only single inheritance. This is a problem if you implement an interface...

vscode-implements-keyword

_Copied from original issue: Microsoft/vscode#3038_

anubmat:
Fixes #89


npm-debug.log
```diff
@@ -0,0 +1,21 @@
...
+8 error node v4.3.1
+9 error npm  v2.14.12
+10 error path C:\Users\anubmat\TypeScript-TmLanguage\package.json
+11 error code ENOENT
+12 error errno -4058
+13 error syscall open
+14 error enoent ENOENT: no such file or directory, open 'C:\Users\anubmat\TypeScript-TmLanguage\package.json'
+14 error enoent This is most likely not a problem with npm itself
+14 error enoent and is related to npm not being able to find a file.
+15 verbose exit [ -4058, true ]
```
DanielRosenwasser: Remove this.

tests/cases/Issue89.ts
```diff
@@ -0,0 +1,4 @@
+export class ^FileLogger ^extends /**/^One ^implements/**/ ^IDisposable{
+	^	
+}
+^
```
zhengbli: Add more tests for complex cases:
1. `extends` more than one things: `export class A extends B, C, D implements E, F, G {`
2. switch the positions of `extends` and `implements`

zhengbli: Also add a marker inside the `/**/` comment block to make sure they are parsed as comment

zhengbli:
:+1:

Issue #96, opened 2016-02-23
'let' appears as keyword inside identifers


aeschli:
_From @f111fei on February 23, 2016 6:14_

version： 0.10.10
![2](https://cloud.githubusercontent.com/assets/7069719/13243197/80630302-da37-11e5-9ccf-a63d6f23433d.png)

`letter`

_Copied from original issue: Microsoft/vscode#3270_

PR #101, opened 2016-02-27
Fix var/let highlighting (issue 96)

sparecycles:
Simple problem, simple fix?

https://github.com/Microsoft/TypeScript-TmLanguage/issues/96

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -11,7 +11,7 @@ patterns:
 repository:
   var-expr:
     name: meta.var.expr.ts
-    begin: (?<!\()\s*\b(var|let|const(?!\s+enum)\b)
+    begin: (?<!\()\s*\b(var|let|const(?!\s+enum))\b
     beginCaptures:
       '1': { name: storage.type.ts }
     end: (?=$|;)
--- a/TypeScript.tmLanguage
+++ b/TypeScript.tmLanguage
@@ -1756,7 +1756,7 @@
 		<key>var-expr</key>
 		<dict>
 			<key>begin</key>
-			<string>(?&lt;!\()\s*\b(var|let|const(?!\s+enum)\b)</string>
+			<string>(?&lt;!\()\s*\b(var|let|const(?!\s+enum))\b</string>
 			<key>beginCaptures</key>
 			<dict>
 				<key>1</key>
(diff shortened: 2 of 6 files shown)


DanielRosenwasser:
Thanks for the PR! From a glance, Can you add this test?

```ts
var ^^constenum = 10;

^^constenum += 10;

^^const ^^enum { }


var ^^constable = "I am rrrrrroyal!";

^^constable += " Royal, I say!"
```

sparecycles:
@DanielRosenwasser I added those tests and verified that the change does not affect those lines (which I believe is correct behavior).

DanielRosenwasser:
Sorry @sparecycles, I didn't realize that the markers checked scopes to the **left**. Can you fix up the test?

```ts
var ^^constenum^^ = 10;

const^^enum^^ += 10;

const^^ enum^^ E { }


var const^^able^^ = "I am rrrrrroyal!";

const^^able^^ += " Royal, I say!"
```

sparecycles:
@DanielRosenwasser I'm not familiar with the test framework either.  Updated to your new request.

Test still differs from prior behavior only in the 3rd and 4th line of code.

And the result looks good :wink:

screen shot 2016-02-28 at 3 15 04 am

DanielRosenwasser:
Beautiful - @zhengbli, care to take a quick look before I pull in?

zhengbli:
Looks good to me :+1:

sparecycles:
@DanielRosenwasser ping.

DanielRosenwasser:
Thanks @sparecycles!

rob3c:
Hi, I see this was merged to master almost a month ago now, but I just updated to typescript 1.9.0-dev.20160328 (which I believe is built from master) and am still seeing the var syntax highlighting problem.

I'm using VSCode 0.10.11 with a custom settings.json file that points to my local install, and VSCode shows the local typescript version is in the status bar in the bottom right. Is there something I'm missing, or is this still not available yet? Or maybe VSCode isn't really using the local version for syntax highlighting? Thanks in advance.

rob3c:
Ah ok - I just found the VSCode Insiders release at https://code.visualstudio.com/insiders. It handles the var/let syntax highlighting correctly for the same nightly version, so it looks like they are handling the highlighting differently beyond the installed typescript version.

PR #102, opened 2016-02-29
Multi Line Strings: Escape character at end of each line in a string allows multi line strings

anubmat:
Solving Issue #5, An escape character at the end of each line in a string allows multi line strings.


zhengbli:
:+1: Seems to be working fine for me.

PR #103, opened 2016-02-29
Fixes the color of the switch keyword.

rbuckton:
We classify `switch` as `keyword.control.ts` in `#control-statement`, but as `keyword.operator.ts` in `#switch-expression`. 

This change modifies `#switch-expression` so that the `switch` keyword is colored as `keyword.control.ts` in both TypeScript and TypeScriptReact syntaxes.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -412,7 +412,7 @@ repository:
     name: switch-expression.expr.ts
     begin: \b(switch)\b\s*\(
     beginCaptures:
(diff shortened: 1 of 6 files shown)


zhengbli:
:+1:

zhengbli:
@rbuckton one test seems to be broken after the merging. Can you take a look? Thanks

anubmat:
@rbuckton , I think you can update the baseline of javascript.ts you added. The one generated looks good.

rbuckton:
Fixed in master.

Issue #37, opened 2016-01-06
operator symbols and keywords all get 'keyword.operator' scope


aeschli:
Try the following example in VS code (0.10.6) in the dark theme:

```typescript
module Body {
    export class Cell {
        public parent: Cell;

        constructor(parent: Cell) {
            this.parent = parent;
            if (typeof parent === 'object' || parent instanceof Error) {
            }
        }
 
```
![image](https://cloud.githubusercontent.com/assets/6461412/12139947/3fa94cca-b464-11e5-8407-46109c781345.png)

Note that `instanceof` and `typeof` are rendered white instead of blue as other keywords.

The TS grammar uses 'keyword.operator' for operator symbols like '=', '+', '!=', but also for JS keywords as switch, instanceof, import, delete, export, import, from, in, new, void, as, of.

We want to colorize the keywords, but not the symbols.

My suggestion is to use 'keyword.operator' only for symbols
 - use 'keyword' for 'instanceof', 'typeof', 'is', 'as', 'in', 'of', 'delete', 'new', 
 - for 'import' I'd suggest to use 'keyword.control.import.include'
 - for 'export' and 'void', I'd suggest 'storage.type'
 - for 'switch' use 'keyword.control'

zhengbli:
So is it correct or not that the `module`, `typeof` and `parent` here has the `keyword.operator.ts` scope? If not, what do you suggest their scopes should be?

aeschli:
I'd suggest to give `module`,`instanceof` and `typeof` just the `keyword.ts` scope (currently: `keyword.operator.ts`)

aeschli:
Same problem with `namespace`

anubmat:
Solved by pull request #105. Closing.

Issue #29, opened 2015-12-15
Yield keyword is not coloured in JavaScript


aeschli:
_From @joaomoreno on December 14, 2015 13:52_

![image](https://cloud.githubusercontent.com/assets/22350/11782623/458b351c-a272-11e5-8bf2-ca1dc66c6560.png)

_Copied from original issue: Microsoft/vscode#1269_

aeschli:
@joaomoreno Please always paste the code snippet as well...

aeschli:
It's a problem of the grammar

aeschli:
_From @joaomoreno on December 15, 2015 10:49_

Sorry!

```javascript
yield
```

:+1:

aeschli:
_From @bpasero on December 15, 2015 10:58_

I think that is not the full snippet from the screenshot...

zhengbli:
Fixed by #117

PR #105, opened 2016-03-01
Solving Issue 37, changing scopes of major keywords, changing scope o…

anubmat:
Solving Issue #37
- Changed scopes of major keywords
- Changed scope of switch statement
- Modified type-annotation to allow 'typeof' keyword

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -339,7 +339,7 @@ repository:
     name: meta.type.function.return.ts
     begin: '=>'
     beginCaptures:
-      '0': { name: keyword.others.ts }
+      '0': { name: keyword.operator.ts }
     end: (?=\s*[,\)\{=;>]|//|$)
     patterns:
     - include: '#type'
@@ -380,7 +380,7 @@ repository:
   variable-initializer:
     begin: (=)
     beginCaptures:
-      '1': { name: keyword.others.ts }
+      '1': { name: keyword.operator.comparison.ts }
     end: (?=$|[,);=])
     patterns:
       - include: '#expression'
@@ -403,6 +403,7 @@ repository:
     - include: '#block'
     - include: '#import-operator'
     - include: '#expression-operator'
+    - include: '#imply-operator'
     - include: '#relational-operator'
     - include: '#arithmetic-operator'
     - include: '#logic-operator'
@@ -495,7 +496,11 @@ repository:
 
   expression-operator:
     name: keyword.others.ts
-    match: =>|\b(delete|in|instanceof|new|typeof|as|is|of)\b
+    match: \b(delete|in|instanceof|new|typeof|as|is|of)\b
+
+  imply-operator:
+    name: keyword.operator.ts
+    match: =>
 
   import-operator:
     name: keyword.control.import.include.ts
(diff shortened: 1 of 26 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -490,8 +492,12 @@ repository:
     - include: '#expression'
 
   expression-operator:
-    name: keyword.operator.ts
-    match: =>|\b(delete|export|import|from|in|instanceof|new|typeof|void|as)\b
+    name: keyword.others.ts
+    match: =>|\b(delete|in|instanceof|new|typeof|as|is|of)\b
```
DanielRosenwasser: Yield should go here

TypeScript.YAML-tmLanguage
```diff
@@ -511,7 +517,7 @@ repository:
 
   storage-keyword:
     name: storage.type.ts
-    match: \b(var|let|function|const|module|namespace)\b
+    match: \b(var|let|function|const|module|namespace|void|export)\b
```
DanielRosenwasser: There are also `void` expressions so I would move those out.

TypeScript.YAML-tmLanguage
Gheoan: Shouldn't `yield` be `control-statement` like in #117 ?

tests/cases/Issue37.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+var m: ^^typeof M = M;
+
+var adder = ^^new Adder(3, 4);
+^^delete (adder)
+
+var s = typeof {}
+interface I {}
+var j: I = {}
```
zhengbli: Can you add the original issue repro text in #37 to this test? namely
```ts
module Body {
    export class Cell {
        public parent: Cell;

        constructor(parent: Cell) {
            this.parent = parent;
            if (typeof parent === 'object' || parent instanceof Error) {
            }
        }
```

TypeScript.YAML-tmLanguage
anubmat: Lets consider the keyword 'yield' as a separate pull request for issue #29.

aeschli:
I tried the changes and they look good to me, with two exceptions: `=>` and `=` for default parameters are now also keyword.other. From a coloring perspective I'd suggest to keep them as keyword.operand.

TypeScript.YAML-tmLanguage
```diff
@@ -339,7 +339,7 @@ repository:
     name: meta.type.function.return.ts
     begin: '=>'
     beginCaptures:
-      '0': { name: keyword.others.ts }
+      '0': { name: keyword.operator.ts }
```
zhengbli: @aeschli suggested to make to [`keyword.operand`](https://github.com/Microsoft/TypeScript-TmLanguage/pull/105#issuecomment-195747412), does it make a difference? Or maybe he meant `operator`

anubmat: There was no field as keyword.operand. And then he said -- keep them as --. 
Considering the definitions of operator and operand I felt he probably meant operator, which they were earlier.

zhengbli:
:+1: 
Also tested using Sublime, the coloring looks good to me.

TypeScript.YAML-tmLanguage
aeschli: Sorry, `keyword.operator`of course, as it was before.

Issue #84, opened 2016-02-11
` breaks TS coloring


aeschli:
_From @isidorn on February 4, 2016 14:29_

* Latest vscode, open debugViewer.ts, notice the coloring is broken after the `
* Today's vscode version broke this, previously it was coloring correctly

![screen shot 2016-02-04 at 15 28 49](https://cloud.githubusercontent.com/assets/1926584/12817770/0462572c-cb54-11e5-9de2-24b78df48445.png)

_Copied from original issue: Microsoft/vscode#2712_

aeschli:
_From @isidorn on February 8, 2016 15:33_

I still see this all the time, adding important label

aeschli:
_From @dbaeumer on February 11, 2016 11:49_

@aeschli did you do any grammar update ?

aeschli:
Yes, I updated the typescript grammars. We're now on the commit from Feb 9.

I can reproduce with this snippet. Seems to be related to the occurrence of `var`, which is highlighted by itself, which is also a bug. 

```typescript
const $ = dom.emmet;
const booleanRegex = /^true|false$/i;
const stringRegex = /^(['"]).*\1$/;

export function renderVariable(tree: tree.ITree, variable: model.Variable, data: IVariableTemplateData, debugInactive: boolean, showChanged: boolean): void {
	if (variable.available) {
		data.name.textContent = `${variable.name}`;
	}

	if (variable.value) {
		renderExpressionValue(variable, debugInactive, data.value, showChanged);
	} else {
		data.value.textContent = '';
		data.value.title = '';
	}
}
```

anubmat:
Looks like it isn't a problem anymore for both vscode and Sublime Text
vscode:
![issue_84_vscode](https://cloud.githubusercontent.com/assets/17298501/13481559/854ff3c4-e09b-11e5-8656-57d3be77c0b5.PNG)
Sublime Text:
![issue_84_sublime](https://cloud.githubusercontent.com/assets/17298501/13481602/df1bf542-e09b-11e5-9773-db61debadbae.PNG)

PR #106, opened 2016-03-02
` breaks TS coloring

anubmat:
Solving Issue #84


DanielRosenwasser:
I don't think this is the correct fix. Let's discuss offline.

Issue #32, opened 2015-12-20
ES6 coloring issues: import, constructor


egamma:
Extracted from Microsoft/vscode#1470 reported by @MacKentoch

Got same issues with **javascript ES6**.
As examples
 -  `import` (was red before) is now just grey
 -  `constructor` is now grey

*Occurs in  `OSX El capitan` and `windows 10`.*

Jsx seems not that much impacted (but still improvements to be done).

MacKentoch:
@egamma @adnan333bd I confirm : 

**Big thanks!**

for 
- fast reply
- accurate analysis 
- and super nice VSCode.

For now, I switched theme and it ok.
Take all the time you need, I will go back dark the when it's ok.

aeschli:
Right now, imports get the `keyword.operator` scope. `operator` is the problem. I'd suggest to change it to `keyword.control.import` as recommended by  http://www.apeth.com/nonblog/stories/textmatebundle.html

For constructors, it probably should also get the `keyword' scope.

aeschli:
`import` is covered by #37

Renaming this issue for 'constructor' which is not colored at all. I would suggest 'support.type'

PR #107, opened 2016-03-02
Fixing scope and color of word "constructor"

anubmat:
Fixing issue #32

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -154,15 +154,16 @@ repository:
 
   method-declaration:
     name: meta.method.declaration.ts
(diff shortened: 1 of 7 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -154,15 +154,16 @@ repository:
...
       '1': { name: storage.modifier.ts }
       '2': { name: storage.modifier.ts }
       '3': { name: storage.modifier.ts }
       '4': { name: storage.type.property.ts }
       '5': { name: keyword.operator.ts }
-      '6': { name: entity.name.function.ts }
-      '7': { name: keyword.operator.ts }
+      '6': { name: support.type.ts }
+      '7': { name: entity.name.function.ts }
+      '8': { name: keyword.operator.ts }
```
DanielRosenwasser: Can you leave a comment on what each of these maps to

zhengbli:
:+1:

aeschli:
@anubmat @zhengbli I think `storage.type` is the better scope (instead of `support.type`)
storage.type describes the type of a element (like var, function, ...)
support.type is a type reference

anubmat:
@aeschli, I will change the scope of 'constructor'  to storage.type with pull request #105. Thanks!

Issue #44, opened 2016-01-11
storage.type used for both keywords and type references


aeschli:
_From @Franziskus1988 on December 23, 2015 10:58_

```typescript
interface TestInterface {
  testvar: string <- "string" colored;
```
```javascript
  testfunc(): string <- "string" not colored;
}
```

_Copied from original issue: Microsoft/vscode#1612_

aeschli:
@Franziskus1988 For what language is that?

aeschli:
I can reproduce in typescript
```typescript
interface TestInterface {
  testvar: string;
  testfunc(): string;
}
```
![image](https://cloud.githubusercontent.com/assets/6461412/12236972/ad24bffc-b87c-11e5-8e12-82054fb74949.png)

The problem is that the `interface` and `the first `string` both get the `storage.type.ts` scope.
We would like to color type names with a different color from the keyword. Maybe give `interface` `keyword` scope instead?

aeschli:
Also `function` and `var` get the `storage.type` scope.
looking at http://tmtheme-editor.herokuapp.com/#!/editor/url/https://raw.githubusercontent.com/voronianski/oceanic-next-color-scheme/master/Oceanic%2520Next.tmTheme
it's probably correct to do so, and all type references should use `entity.name.type.class`

zhengbli:
Closing as fixed by #108

PR #108, opened 2016-03-03
Solving Issue 44 - Changed scope name for type references

anubmat:
Solving Issue #44,
- using ‘storage.type’ only for JS keywords like function, var, let, const
- using ‘support.type‘ for the type declaration


DanielRosenwasser:
:+1:

Issue #38, opened 2016-01-06
'implements' not correctly tokenized


aeschli:
Try the following example in VS code (0.10.6) in the dark theme:

```typescript
module Conway {
    export interface Foo {
    }
    export class Bar {
    } 
    export class Cell extends Bar implements Foo {
    }
}
```
![image](https://cloud.githubusercontent.com/assets/6461412/12140156/eef8783a-b465-11e5-9eeb-aeb99cb8a547.png)

Note that `implements` is colored like the type. It gets the same scope as the type name
(`block declaration heritage meta object parent storage type ts`)

PR #109, opened 2016-03-03
Correctly classify keywords in object-heritage clauses

anubmat:
Solving Issue #38 / #89. 
Improving upon the previous pull request #99

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -89,7 +89,7 @@ repository:
 
   object-heritage:
     name: meta.object.heritage.ts
-    begin: '(?:\b(extends|implements)\b)'
+    begin: '(?:\s*\b(extends|implements)\b)'
     beginCaptures:
       '1': { name: keyword.other.ts }
     end: (?=\{)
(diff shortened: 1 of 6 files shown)


DanielRosenwasser:
In what way is this an improvement?

anubmat:
The earlier fix was insufficient. We looked for word boundaries but not spaces. 
Example:
Regexp  "{\bname" matches 
take1
but does not match
take2
However, Regexp "{\s*\bname" matches both.

Spaces before the word 'implement' after 'extends' or vice-versa were not being matched/consumed by any of the patterns (Cant say why they are not being ignored)

zhengbli:
This is a little weird as I think `(?:\b(extends|implements)\b)` should cover the same cases with `(?:\s*\b(extends|implements)\b)`. I'm guessing the reason why they behave differently was that some other pattern may match the spaces and then takes the chance for this pattern to function. 

Can you add to the test what cases that wasn't working after #99 but was fixed by this PR? That way we'll make sure the test is comprehensive.

anubmat:
Thanks @zhengbli. Your guess was right. There were spaces in the beginning of object-heritage parent which made it difficult for regexp to match. Output looks right too!

zhengbli:
:+1:

Issue #110, opened 2016-03-04
numbers in exponential notation not correctly colored.


aeschli:
_From @wuyiw on March 4, 2016 11:26_

numbers in exponential notation like 1e3, 3600e3... are not colored as normal number: 
[http://i.imgur.com/LfnxtiA.png](http://i.imgur.com/LfnxtiA.png)

_Copied from original issue: Microsoft/vscode#3710_

PR #118, opened 2016-03-10
Allowing numbers in exponential notation to be scoped as numeric literals

anubmat:
Solving Issue #110, numbers in exponential notation not colored

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -631,7 +631,7 @@ repository:
 
   numeric-literal:
     name: constant.numeric.ts
-    match: \b(?<=[^$])((0(x|X)[0-9a-fA-F]+)|([0-9]+(\.[0-9]+)?))\b
+    match: \b(?<=[^$])((0(x|X)[0-9a-fA-F]+)|(([0-9]+(\.[0-9]+)?))(e[0-9]+(\.[0-9]+)?)?)\b
 
   boolean-literal:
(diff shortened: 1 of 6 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -631,7 +631,7 @@ repository:
 
   numeric-literal:
     name: constant.numeric.ts
-    match: \b(?<=[^$])((0(x|X)[0-9a-fA-F]+)|([0-9]+(\.[0-9]+)?))\b
+    match: \b(?<=[^$])((0(x|X)[0-9a-fA-F]+)|(([0-9]+(\.[0-9]+)?))(e[0-9]+(\.[0-9]+)?)?)\b
```
zhengbli: Capital `E` as well

tests/cases/Issue110.ts
```diff
@@ -0,0 +1,10 @@
+let number1 = ^^0x100   
+let number2 = ^^14e15
+let number3 = ^^10.42e41.5 
+let number4 = ^^4
+let number5 = ^^51.4
+
```
zhengbli: Also try to add the original repro code in #110 to see what if the exponential notation is the second assignment in the same line

zhengbli: Add a test case for the `+` sign, e.g.
```
let number6 = 5.00567789e+2
```

zhengbli:
:+1:

Issue #81, opened 2016-02-10
JS ternary expression not colored correctly


Tyriar:
Under vscode master (0.10.9) with language mode JavaScript.

![ternary](https://cloud.githubusercontent.com/assets/2193314/12963913/0eee1674-d002-11e5-8d01-08f8609b26e1.png)

`'amd64'` has classes: token block decl expr function meta single string ts var var-single-variable js
`'i386'` has classes: token **annotation** block decl expr function meta single string ts type var var-single-variable js

`annotation` scope is causing the following theme style to be applied:

```css
.monaco-editor.vs-dark.vscode-theme-colorful-defaults-themes-dark_plus-tmTheme .token.var.annotation.type.js {
    color: rgba(78, 201, 176, 1);
}
```

thorn0:
Still happens with vscode 0.10.10.

SamVerschueren:
Should the ternary operator be declared on itself? I searched through the document but couldn't find a direct match to be honest.

Tyriar:
You generally only scope something if it's useful for styling wouldn't you? I would expect the `?` and `:` to be scoped as symbols or operators?

SamVerschueren:
Does this mean there is a false-positive match which adds the `annotation` style? Not very familiar with how this all works. Trying to get a grip on it :).

Tyriar:
I'm not sure what annotation is supposed to me, some TypeScript syntax related to `:`? I do notice however that removing the `var` fixes the issue so this seems specific to declaring and initializing variables with ternary operators.

![image](https://cloud.githubusercontent.com/assets/2193314/15584713/9ba827a4-2330-11e6-9e3e-49cd547c5b02.png)

zhengbli:
Fixed by #129

Issue #80, opened 2016-02-09
Can't handle string template in ternary


blakeembrey:
![image](https://cloud.githubusercontent.com/assets/1088987/12932771/67f1cc94-cf39-11e5-9ed6-80551ff12a41.png)

anubmat:
Duplicate #124

sheetalkamat:
Fixed by #257

PR #129, opened 2016-03-17
Adding capability for ternary expression as a separate scope

anubmat:
Solving Issue #124, #80, #81

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -26,9 +26,16 @@ repository:
       '1': { name: variable.ts }
     end: (?=$|[;,])
     patterns:
+    - include: '#ternary-operator'
     - include: '#type-annotation'
     - include: '#string'
     - include: '#comment'
+    - include: '#expression'
+
+  ternary-operator:
+    begin: (\?)
+    end: (:)
+    patterns:
     - include: '#expression'
 
   control-statement:
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -26,17 +26,44 @@ repository:
       '1': { name: variable.ts }
     end: (?=$|[;,])
     patterns:
-    - include: '#ternary-operator'
+    - include: '#ternary-expression'
     - include: '#type-annotation'
     - include: '#string'
     - include: '#comment'
     - include: '#expression'
 
+  ternary-expression:
+    begin: (?=\?)
+    end: (?=$|[;,])
+    patterns:
+    - include: '#ternary-operator'
+    - include: '#ternary-expression-type' 
+
   ternary-operator:
     begin: (\?)
     end: (:)
     patterns:
-    - include: '#expression'
+    - include: '#ternary-expression-type' 
+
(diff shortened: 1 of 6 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -26,11 +26,18 @@ repository:
       '1': { name: variable.ts }
     end: (?=$|[;,])
     patterns:
+    - include: '#ternary-operator'
```
DanielRosenwasser: Why isn't `ternary-operator` just part of `expression`?

TypeScript.YAML-tmLanguage
```diff
@@ -26,11 +26,18 @@ repository:
...
     - include: '#type-annotation'
     - include: '#string'
     - include: '#comment'
     - include: '#expression'
 
+  ternary-operator:
```
DanielRosenwasser: These are ternary *expressions*, not operators

anubmat: I will make that change. Thanks!

TypeScript.YAML-tmLanguage
anubmat: If what you mean is why we need a 'ternary-operator' as a new addition, there are 2 reasons I thought of:
1) To make (? + expression + :) together
2) #Type-annotation consumes " : followed by something" before we can get to #expression. Even if we change the order in which they are present. It will include : followed by something as #type-annotation instead of skipping it to look for an expression.

TypeScript.YAML-tmLanguage
anubmat: ('?' and ':') together are known as the ternary operator, not sure if we should call it expression or operator. Also note that we end at ':', which can be followed by an expression (since expressions can exist anywhere.)

billti: Is it safe just to end there? I know an expression can exist anywhere, but not any production can come after the ":". For example, is this now parsed as a code block (with a labeled statement) after the ":" instead of an expression:

```javascript
var x = true ? {a: 42} : {label: "use strict"};
```

anubmat: Agreed, it is not safe to end here since even in other scenarios like the following, highlighting shows no issues, though it should. 
```typescript
var x = true ? {a: 42} : {b: 43};
var y = true ? {a: 42} : let var2 = 8;
```
I will add a bigger scope called ternary-expression that only allows this ternary-operator followed by an expression, and see if that works. Thanks.

TypeScript.YAML-tmLanguage
```diff
@@ -26,11 +26,45 @@ repository:
...
+    - include: '#ternary-operator'
+    - include: '#ternary-expression-type' 
+
+  ternary-operator:
+    begin: (\?)
+    end: (:)
+    patterns:
+    - include: '#ternary-expression-type' 
+
+  ternary-expression-type:
```
DanielRosenwasser: Can you document (in comments) exactly what this is supposed to capture? For instance, why do you need types covered here?

zhengbli:
:+1:

Issue #136, opened 2016-03-26
JavaScript highlighting broken after ternary expression with template strings with place holders


shamrin:
Highlighting is broken after ternary expression with template strings, with place holder in it:

```js
let a = true ? c : `${e}`;
let f = 1;
```

Here's how VSCode renders it (`let f = 1` is no longer highlighted):

![1](https://cloud.githubusercontent.com/assets/510678/14061995/0c1d5138-f3a9-11e5-9c9b-fabd8d68a415.png)

This problem makes VSCode (which is otherwise wonderful!) unusable in the project I work on.

Visual Studio Code
Version 0.10.11 (0.10.11)
f291f4ad600767626b24a4b15816b04bee9a3049
2016-03-11T15:23:23.736Z

Different, but probably related issue: #120

anubmat:
Duplicate #124. Closing here.

PR #137, opened 2016-03-26
Add test for issue #136 (currently fails)

shamrin:
This adds a test for issue #136 (but doesn't fix the bug yet).


anubmat:
Since this pull request does not solve the issue yet, the baselines generated may not be correct. This test case will be added when the issue is solved.  Thanks

PR #145, opened 2016-04-18
Parse longer expressions first, to help ligatures in Editors

long-lazuli:
This helps with this issue in atom-typescript : https://github.com/TypeStrong/atom-typescript/issues/851
which I first wanted to fix here : https://github.com/TypeStrong/atom-typescript/pull/926


zhengbli:
:+1: LGTM. Thanks!

PR #179, opened 2016-06-07
Attempt to fix #154

### LINKED ISSUE #154: String method names do not terminate string syntax highlighting
_From @chadhietala on April 23, 2016 20:57_

- VSCode Version: 1.0.0 
- OS Version: OSX 10.10.5

Steps to Reproduce:

1. Have a class with stings as method names
2. Observe syntax highlighting is different from non-string method names

__String Method Name__
screen shot 2016-04-23 at 1 56 50 pm

__Method Name__
screen shot 2016-04-23 at 1 55 35 pm

_Copied from original issue: Microsoft/vscode#5703_

chadhietala:
I actually have no idea if this is correct or not but thought I would put it up to get some feedback.

____

***Edit by @DanielRosenwasser:*** *Fixes #154.*


DanielRosenwasser:
Hey @chadhietala, as a heads up, mentioning the issue you're fixing in the PR body actually links to it, and the PR title itself should typically be more descriptive so readers don't have to navigate to the original issue for full context.

@anubmat can you take a look at this?

chadhietala:
Not sure how important the CLA is now that I will be working at MS subsidiary... 😄

.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 // Place your settings in this file to overwrite default and user settings.
 {
-    "editor.insertSpaces": true
-}
\ No newline at end of file
+    "editor.insertSpaces": true,
+    "files.trimTrailingWhitespace": false
```
zhengbli: Why this change?

TypeScript.YAML-tmLanguage
```diff
@@ -154,15 +154,15 @@ repository:
 
   method-declaration:
     name: meta.method.declaration.ts
-    begin: '\b(?:(abstract)\s+)?\b(?:(public|private|protected)\s+)?\b(?:(async)\s+)?(?:(get|set)\s+)?(?:(new)|(?:\b(constructor)\b)|(?:([a-zA-Z_$][\.\w$]*)\s*(\??)))?\s*(?=\(|\<)'
+    begin: '\b(?:(abstract)\s+)?\b(?:(public|private|protected)\s+)?\b(?:(async)\s+)?(?:(get|set)\s+)?(?:(new)|(?:\b(constructor)\b)|(?:(\[*)(`|'|"*)([a-zA-Z_$][\.\w$]*)(`|'|"*)(\]*)\s*(\??)))?\s*(?=\(|\<)'
```
zhengbli: The thing inside the ``|'|"` could be anything, not only `[a-zA-Z_$][\.\w$]*`

For example ``[`123`]``, ``[`@$!$`]`` are both valid method name

.vscode/settings.json
chadhietala: I can remove this. I think I'm using [this extension](https://marketplace.visualstudio.com/items?itemName=samverschueren.final-newline) that adds a new line and tests care about whitespace.

TypeScript.YAML-tmLanguage
```diff
@@ -154,15 +154,15 @@ repository:
 
   method-declaration:
     name: meta.method.declaration.ts
-    begin: '\b(?:(abstract)\s+)?\b(?:(public|private|protected)\s+)?\b(?:(async)\s+)?(?:(get|set)\s+)?(?:(new)|(?:\b(constructor)\b)|(?:([a-zA-Z_$][\.\w$]*)\s*(\??)))?\s*(?=\(|\<)'
+    begin: '\b(?:(abstract)\s+)?\b(?:(public|private|protected)\s+)?\b(?:(async)\s+)?(?:(get|set)\s+)?(?:(new)|(?:\b(constructor)\b)|(?:(\[*)(`|'|"*)(.*)(`|'|"*)(\]*)\s*(\??)))?\s*(?=\(|\<)'
```
zhengbli: This still doesn't look correct. For example, the `.*` would eat everything including the closing quote and bracket (the test case needs to be updated for it). Also, do you need to consider if the two quotes are matching? For example, `["test']` and `['test"]` shouldn't be treated the same way as the `["test"]`

tests/cases/Issue154.ts
```diff
@@ -0,0 +1,35 @@
+class TestCase0 {
+^^['123']() {
```
zhengbli: You should also test if the rule ends correctly. e.g.
```
^^['123']^^()
```

mhegazy:
The original issue should be fixed now by https://github.com/Microsoft/TypeScript-TmLanguage/pull/257

PR #199, opened 2016-06-30
Allow method return types to end with a digit

gcnew:
Syntax coloring fails if method return type ends with a digit. This pull request allows digits to be the last character in `type-annotation.end` regular expression.

Fixes https://github.com/Microsoft/TypeScript/issues/9444 and should be in line with the to-be-added numeric literal types https://github.com/Microsoft/TypeScript/pull/9407.

### DIFF
--- a/tests/cases/FunctionMethodReturnTypes.ts
+++ b/tests/cases/FunctionMethodReturnTypes.ts
@@ -28,7 +28,7 @@ class TestClass {
     ^^public ^^testMethodReturnType11(): number | string { ^^return 1 }
     ^^public ^^testMethodReturnType12(): number | string [] { ^^return }
     ^^public ^^testMethodReturnType13(): [number, number] | string [] { ^^return [""] }
-    ^^public ^^testMethodReturnType14(): EndsWithDigit1 { ^^return 123 }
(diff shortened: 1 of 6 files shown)


tests/cases/FunctionMethodReturnTypes.ts
```diff
@@ -28,4 +28,7 @@ class TestClass {
     ^^public ^^testMethodReturnType11(): number | string { ^^return 1 }
     ^^public ^^testMethodReturnType12(): number | string [] { ^^return }
     ^^public ^^testMethodReturnType13(): [number, number] | string [] { ^^return [""] }
-}
\ No newline at end of file
+    ^^public ^^testMethodReturnType14(): EndsWithDigit1 { ^^return 123 }
```
anubmat: Can you add a check to scope of 123 and after 123, like -- 
```ts
^^return ^^123 ^^} 
```

gcnew: Done.

mhegazy:
Sorry for the delay, this should be fixed now by https://github.com/Microsoft/TypeScript-TmLanguage/pull/257

PR #201, opened 2016-07-06
added: ES7 deocrators (angular 2), html syntax highlight in template string

orizens:
hi. 
These updated add support for decorators syntax highlighting and html syntax in template strings. 
As a reference: 
* I added html support by following [this thread ](https://forum.sublimetext.com/t/javascript-es6-template-literals-syntax-for-html/18242/2) 
* I added es7 decorators support by inspecting the Babel yaml language definition filw

screen shot 2016-07-06 at 12 19 50 pm


DanielRosenwasser:
Hey @orizens. I think we should close these and start completely fresh on an implementation. Let's also keep the scope small (i.e. no HTML highlighting in templates). This should just cover decorators.

Here's a reasonable amount I think we'd expect in the implementation:

* Assume that a decorator can occur anywhere in the file.
* Assume that a decorator starts with an `@` and an arbitrarily long series of dotted names (i.e. `@foo.bar.baz` or `@foo`), which all occur on the same line.
* Assume that decorators come in two flavors - they are either
    * plain member expression decorators (i.e. `@foo.bar`) or
    * call expression decorators (i.e. `@foo.bar("Hello world!")`).
* Since these member/call expressions are still expressions, make sure you use an appropriate capture group for them.
* Any amount of whitespace can come after the `@`, after the `foo.bar`, and before/after each dot.

If you think you'd be up to a clean implementation, we could do that. If not, let us know and we could have the team take a swing at it.

orizens:
Hi @DanielRosenwasser 
i'll make a separate pr for the html syntax.
As for the decorators - it'll be better if your team take it. 
Thanks.

DanielRosenwasser:
Sure - for the record, I don't think we should classify template strings blindly as HTML. If it's possible, we should look at common places where we know a template is intended to be HTML. Angular templates are one place, so if you see a property named `template` and a template string, then consider highlighting that as HTML.

orizens:
@DanielRosenwasser 
I see your point. we can narrow the regex look up to include 'template' - is this the direction you're implying? 

However, if there's no html syntax inside a template string - it is regarded simply as a string (from a user perspective). 
Consider html code can be used elsewhere other than the "template" property - when typescript is not written for an angular2 project. 
All in all, as a consumer of Typescript, where I currently use it to write angular2 - i'de like to have html syntax highlight in template strings - the ux as a developer is much needed and improves readability - and that's a feature that in my opinion, should be regarded. 
Thanks.

dgaleazzo:
@orizens I'm using your commit in sublime text and it works well! Thank you, I hope that you will work to pass all test units in order to implement it officially.

anubmat:
@orizens, thanks for opening a separate PR. can you please update React files as well?

orizens:
@anubmat 
updating React files with both Decorator and Html String Templates support?

anubmat:
Yes, whenever you update tmLanguage, you update both YAMLs - 4 files should change at least.
I am a little confused here - This PR has both decorators and HTML highlighting stuff. You said you did not want to consider working on decorators in non-Babel way. If that is the case, can you please update this PR to only aim at HTML highlighting and remove the decorators part or alternatively open another PR.

orizens:
@anubmat 
I'm closing this PR and open a PR for HTML syntax only. 
Thanks.

orizens:
new PR #220

anubmat:
Thanks @orizens

PR #209, opened 2016-07-15
Fixes Issue #200, Correcting case/default expression clause

anubmat:
Fixes issue #200 
The expression clause in switch case-clause and default clause included recursive references to expression scope.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -37,15 +37,15 @@ repository:
     end: (?=$|[;,])
     patterns:
     - include: '#ternary-operator'
-    - include: '#ternary-expression-type' 
+    - include: '#expression-type' 
 
   ternary-operator:
     begin: (\?)
     end: (:)
     patterns:
-    - include: '#ternary-expression-type' 
+    - include: '#expression-type' 
 
-  ternary-expression-type:
+  expression-type:
     name: meta.expression.ts
     patterns:
     - include: '#string'
@@ -466,12 +466,12 @@ repository:
 
   case-clause:
     name: case-clause.expr.ts
-    begin: '(?<!\.)\b(case|default)\b(?=:)'
+    begin: '(?<!\.)\b(case|default)\b'
     beginCaptures:
       '1': { name: keyword.control.ts }
     end: ':'
     patterns:
-    - include: '#expression'
+    - include: '#expression-type'
 
   switch-statement:
     name: switch-statement.expr.ts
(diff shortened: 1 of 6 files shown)


anubmat:
@zhengbli , @DanielRosenwasser , can you please review this PR?

TypeScript.YAML-tmLanguage
```diff
@@ -37,15 +37,15 @@ repository:
...
 
   ternary-operator:
     begin: (\?)
     end: (:)
     patterns:
-    - include: '#ternary-expression-type' 
+    - include: '#expression-type' 
 
-  ternary-expression-type:
+  expression-type:
```
DanielRosenwasser: II still don't understand why this is called `expression-type`. Can you give this a more accurate name?

TypeScript.YAML-tmLanguage
```diff
@@ -466,12 +466,12 @@ repository:
 
   case-clause:
     name: case-clause.expr.ts
-    begin: '(?<!\.)\b(case|default)\b'
+    begin: '(?<!\.)\b(case|default(?=:))\b'
```
DanielRosenwasser: What is the precedence here? Is this parsed as "`case` OR `default(?=:)`"? Do you only highlight `case` if it immediately followed by a colon? Can you test this out explicitly?

TypeScript.YAML-tmLanguage
anubmat: Can you suggest something. I do not like expression-type myself. Maybe calling this expression and the other expression  something like expression-group?

TypeScript.YAML-tmLanguage
anubmat: This means case or (default followed by a lookahead for colon). It does not catch "case:".

anubmat: I have tested "export default" which was causing the problem. There are existing tests with switch statement using case and default - and those work as expected.

TypeScript.YAML-tmLanguage
DanielRosenwasser: This is only used in the true/false branch of a ternary expression, right? How about `ternary-expression-branch`?

anubmat: Nopes, I just used it in case-clause because the expression in case clause had a recursive call for expression scope.

anubmat: How about just expression-branch?

DanielRosenwasser: The real problem is that switch statements are not expressions (nor are control statements). But I guess this is fine just to break the circularity.

DanielRosenwasser:
:+1:

anubmat:
Merging this in. Will discuss and fix expression scope-name in a later PR.

PR #210, opened 2016-07-15
Fix CI

zhengbli:


anubmat:
👍

anubmat:
Is this simply an upgrade or does it also fix the build fail issue?

zhengbli:
It should fix it. The problem is that we should use `npm@next` for building node_gyp, see https://github.com/nodejs/node-gyp/issues/972#issuecomment-229887765

PR #220, opened 2016-07-25
[added]: html syntax highlight in template strings

orizens:
hi. 
These updated add support for html syntax in template strings. 
This feature support is useful for several use cases: 

1. if you're developing with Angular 2 using Typescript - it is easier to see a component's template inside a ts file (as shown in the screenshot below). 
2. There are use cases where an html template is constructed with template strings and template variables - it is very helpful to see the markup inside a ts file in this case as well. 

As a reference: 
I added html support by following [this thread](https://forum.sublimetext.com/t/javascript-es6-template-literals-syntax-for-html/18242/2) 

html

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -618,6 +618,7 @@ repository:
     end: (?=(\$\{|`))
     patterns:
     - include: '#string-character-escape'
+    - include: 'text.html.basic'
 
   string-character-escape:
     name: constant.character.escape
--- a/TypeScript.tmLanguage
+++ b/TypeScript.tmLanguage
@@ -1414,6 +1414,10 @@
 					<key>include</key>
 					<string>#string-character-escape</string>
 				</dict>
+				<dict>
+					<key>include</key>
+					<string>text.html.basic</string>
+				</dict>
 			</array>
 		</dict>
 		<key>template-substitution-element</key>
--- a/TypeScriptReact.YAML-tmLanguage
+++ b/TypeScriptReact.YAML-tmLanguage
@@ -607,6 +607,7 @@ repository:
     end: (?=(\$\{|`))
     patterns:
     - include: '#string-character-escape'
+    - include: 'text.html.basic'
 
   string-character-escape:
     name: constant.character.escape
--- a/TypeScriptReact.tmLanguage
+++ b/TypeScriptReact.tmLanguage
@@ -1780,6 +1780,10 @@
 					<key>include</key>
 					<string>#string-character-escape</string>
 				</dict>
+				<dict>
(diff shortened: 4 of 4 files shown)


anubmat:
Looks good to me. @DanielRosenwasser , as discussed before, are there any issues with supporting HTML syntax in template strings. @orizens, can you support this PR with an example why this may be useful.

orizens:
@anubmat I updated the pr description.

DanielRosenwasser:
@orizens would it be possible to move this rule into a separate extension? The way I see it working would be for it to take advantage of the TypeScript template scope and ride on top of that. I get that this is great for the Angular template case, but template strings are not only meant for HTML.

orizens:
hi @DanielRosenwasser 
what do you mean by a separate extension? I'm not sure how a syntax highlight can applied in sublime in addition to another. if you can reference anything on that - it can be helpful. 
This addition doesn't stop the other patterns lookup in template strings - meaning, parsing for template strings variable is processed as expected when no in html.

sheetalkamat:
@orizens look at the information on injection grammar
http://blog.macromates.com/2012/injection-grammars-project-variables/

This change shouldn't be here but an extension with injection grammar rule.

orizens:
@sheetalkamat Thanks for the reference. 
The solution i chose is: 
Finally, I created a new syntax file from Typescript (a menu item in sublime) and added the html syntax definition to the new file. then, i set sublime to open all ts files with this custom-typescript  syntax file. 
simple enough and works.
finally, I saved it as my custom-typescript syntax file and

Fettah:
@orizens  I could not understand what you have written. could you please elaborate more? I also need the syntax highlighter for the html inside the .ts file

youssmak:
@orizens could you please share your syntax file to highlight correctly html and angular 2 directives in .ts files ?

sheetalkamat:
You can try [vscode angular2-inline extension](https://marketplace.visualstudio.com/items?itemName=natewallace.angular2-inline). If you are trying to write syntax file that uses injection grammar for other editors, [here](https://github.com/natewallace/angular2-inline/blob/master/syntaxes/template.ng.json#L3) is the syntax file from this extension

orizens:
@youssmak all relevant files are in this PR https://github.com/Microsoft/TypeScript-TmLanguage/pull/220/files

youssmak:
@orizens thank you, i already see your PR.
I saw you were working with angular and i thought you had a better syntax file for highlight angular 2 directives.

@sheetalkamat i prefer Sublime Text. I tried atom but it's finally to slow compared to ST.
I'm not comfortable with VSCODE UI/UX but i will take a second look at it soon.

Issue #36, opened 2016-01-05
Highlight ES6 object method shorthand syntax


LPGhatguy:
ES6 has a shorthand syntax for defining methods on objects similar to that of a class.

The method declaration in
```ts
let x = {
	hello() {
		let hello = "world";
	}
};
```

should have the same highlighting as
```ts
class X {
	hello() {
		let hello = "world";
	}
}
```

In the current TypeScript Sublime Text plugin, the method name in the first code sample is not highlighted: 
![screen shot 2016-01-05 at 2 44 43 pm](https://cloud.githubusercontent.com/assets/654598/12128513/05f6e8ec-b3bb-11e5-921b-8b0312009327.png)

(as a side note, GitHub has the same highlighting issue at present; do they use this repository?)

sheetalkamat:
Fixed by #257

PR #224, opened 2016-07-28
Allows method declaraions inside Blocks, 

anubmat:
Solves Issue #36


mhegazy:
Sorry for the delay. The original issue should be fixed now by https://github.com/Microsoft/TypeScript-TmLanguage/pull/257

PR #225, opened 2016-07-31
First line matching for node added....

pmkary:


anubmat:
You should generate and update the corresponding plist files

TypeScript.YAML-tmLanguage
```diff
@@ -5,6 +5,8 @@ scopeName: source.ts
 fileTypes: [ts]
 uuid: ef98eb90-bf9b-11e4-bb52-0800200c9a66
 
+- firstLineMatch: #!/.*\bnode
+
```
anubmat: Please update corresponding plist files.

PR #240, opened 2016-08-12
Allowed digits at the end of method return type

### LINKED ISSUE #153: TypeScript syntax highlighting breaks in case the return type of a function ends with a number
_From @Lunatix89 on April 23, 2016 14:43_

- VSCode Version: 1.0.0
- OS Version: Windows 8.1 and Windows 10

Steps to Reproduce:

1. Create a typescript file
2. Insert the following code: 

```typescript
import {Vector3, Vector} from 'vectors';

/**
 * A simple 3x3 matrix structure.
 * 
 * @export
 * @class Matrix3x3
 */
class Matrix {
	public M11 : number;
	public M12 : number;
	public M13 : number;
	public M21 : number;
	public M22 : number;
	public M23 : number;
	public M31 : number;
	public M32 : number;
	public M33 : number;
	
	
	/**
	 * Gets a column as a new vector.
	 * 
	 * @param {number} index The index of the column (0 .. 2).
	 * @returns {Vector3} A vector representing the column.
	 */
	public getColumn_Breaks(index : number) : Vector3 {
		if (index == 0) {
			return new Vector3(this.M11, this.M21, this.M31);
		} else if (index == 1) {
			return new Vector3(this.M12, this.M22, this.M32);
		} else if (index == 2) {
			return new Vector3(this.M13, this.M23, this.M33);
		} else {
			throw new RangeError('Invalid matrix 3x3 column index: ' + index);
		}
	}
	
	/**
	 * Gets a column as a new vector.
	 * 
	 * @param {number} index The index of the column (0 .. 2).
	 * @returns {Vector3} A vector representing the column.
	 */
	public getColumn_Works(index : number) : Vector {
		if (index == 0) {
			return new Vector(this.M11, this.M21, this.M31);
		} else if (index == 1) {
			return new Vector(this.M12, this.M22, this.M32);
		} else if (index == 2) {
			return new Vector(this.M13, this.M23, this.M33);
		} else {
			throw new RangeError('Invalid matrix 3x3 column index: ' + index);
		}
	}
	
}
```

image

Fun fact: Seems even the GitHub highlighter can not handle that specific case :|

image

_Copied from original issue: Microsoft/vscode-tslint#46_

### LINKED ISSUE #188: TypeScript syntax highlighting breaks for class methods returning a type ending in a digit
_From @lostfictions on June 3, 2016 7:45_

This is a bit of a strange one. In the screenshot below, a class with two methods is declared. The first returns a `number` and has correct syntax highlighting. The second has an almost identical body but returns an `endsWithDigit1` (a type alias for `number`), which breaks highlighting for the method body.

(On a more practical level, I discovered this while trying write a method that returned a `Matrix4x4`, so it's definitely something that might happen in day-to-day use!)

bug

- VSCode Version: 1.1.1
- OS Version: Ubuntu 14.04

Steps to Reproduce:

The repro code in the screenshot above is as follows:

```typescript
type endsWithDigit1 = number

class Repro {
  normalMember() : number {

    const someDeclaration : (name : string) => string = (name) => 'hello there ' + name
    const myString = someDeclaration('human')
    let coolNumber = myString.length + 20
    var anotherCoolThing = {
      pigs: 6,
      rabbits: 29,
      horseName: 'Johnny'
    }
    
    console.log('this is how the syntax highlighting should look!')
    
    return 1
  }
  
  badHighlightingMember() : endsWithDigit1 {
    
    const someDeclaration : (name : string) => string = (name) => 'hello there ' + name
    const myString = someDeclaration('human')
    let coolNumber = myString.length + 20
    var anotherCoolThing = {
      pigs: 6,
      rabbits: 29,
      horseName: 'Johnny'
    }
    
    console.log('this has some highlighting but the colors are wrong')
    
    return 1
  }
}
```

_Copied from original issue: Microsoft/vscode#7170_

RichieAHB:
Fixes #153 - includes tests. The regex for `type-annotation` was looking for `[a-zA-Z_$]` (the regex used for the _first_ character) as the pattern for the last character rather than `[\w$]`. There are a few regexes around that allows `[.\w$]` for the last character of a type name but as far as I can tell, type names can't end with a dot so I've not included this.

I have a plugin that removes whitespace at the end of lines so there are some rogue changes in here that I missed, which can be removed if required.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -37,13 +37,13 @@ repository:
     end: (?=$|[;,])
     patterns:
     - include: '#ternary-operator'
-    - include: '#expression-type' 
+    - include: '#expression-type'
 
   ternary-operator:
     begin: (\?)
     end: (:)
     patterns:
-    - include: '#expression-type' 
+    - include: '#expression-type'
 
   expression-type:
     name: meta.expression.ts
@@ -63,7 +63,7 @@ repository:
     - include: '#logic-operator'
     - include: '#assignment-operator'
     - include: '#type-primitive'
-    - include: '#function-call' 
+    - include: '#function-call'
 
   control-statement:
     name: keyword.control.ts
@@ -307,7 +307,7 @@ repository:
   type-annotation:
     name: meta.type.annotation.ts
     begin: ":"
-    end: (?=$|[,);\}\]]|//)|(?==[^>])|(?<=[\}>\]\)]|[a-zA-Z_$])\s*(?=\{)
+    end: (?=$|[,);\}\]]|//)|(?==[^>])|(?<=[\}>\]\)]|[\w$])\s*(?=\{)
     patterns:
     - include: '#expression-operator'
     - include: '#type'
--- a/TypeScript.tmLanguage
+++ b/TypeScript.tmLanguage
@@ -1604,7 +1604,7 @@
 			<key>begin</key>
 			<string>:</string>
 			<key>end</key>
-			<string>(?=$|[,);\}\]]|//)|(?==[^&gt;])|(?&lt;=[\}&gt;\]\)]|[a-zA-Z_$])\s*(?=\{)</string>
+			<string>(?=$|[,);\}\]]|//)|(?==[^&gt;])|(?&lt;=[\}&gt;\]\)]|[\w$])\s*(?=\{)</string>
 			<key>name</key>
 			<string>meta.type.annotation.ts</string>
 			<key>patterns</key>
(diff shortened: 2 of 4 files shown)


anubmat:
Hi @RichieAHB , this PR is similar to PR #199 . I see you have a good test here. You can either leave some comments on PR #199, or wait for that PR to be merged in and send out a PR to improve those changes.

mhegazy:
Sorry for the delay. the original issue should be fixed now by https://github.com/Microsoft/TypeScript-TmLanguage/pull/257

PR #257, opened 2016-09-08
Rewrite and fixes for lot of grammar issues

sheetalkamat:


sheetalkamat:
merging this as suggested by @mhegazy

Tyriar:
Nice work! Looking forward to testing it out.

whitecolor:
When it will be available?

whitecolor:
Should it also effect on this?
https://github.com/Microsoft/vscode/issues/11398

And when will it be available?

mhegazy:
> And when will it be available?

we should have this in a vscode extension for users to try it out next week. if every thing looks good, should be available in vscode's subsequent releases.

octref:
@mhegazy 

Thanks a lot for the hard work! Can you give an overview of the most important grammar issues that get fixed in this PR?

mhegazy:
there are some 120 issues that are fixed by this change (linked above). it is almost a rewrite at this point. give it a try and let us know if there are other issues we are not aware of.

whitecolor:
> give it a try and let us know if there are other issues we are not aware of.

So it is not available yet right? What will be need to it out, just use vscode insiders?

sheetalkamat:
@whitecolor the issue you mentioned seems to be fixed.

farfromrefug:
@sheetalkamat is this going to be merged soon? Can i test it already?
Thanks

mhegazy:
@farfromrefug you can find these changes in the VSCode extension: https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar

farfromrefug:
Thanks @mhegazy will test it right away!

PR #258, opened 2016-09-08
Add primitive types null, true, and false

poke:
Starting with TypeScript 2, we can use all kinds of literal types. These three are very common literal types which also happen to be reserved words, so I guess it makes sense to handle them in a special way.

Ultimately, it might make sense to expand the syntax structure for literal types a bit but for now, this is a quick fix.

### DIFF
--- a/TypeScript.tmLanguage
+++ b/TypeScript.tmLanguage
@@ -1850,7 +1850,7 @@
 				</dict>
 			</dict>
 			<key>match</key>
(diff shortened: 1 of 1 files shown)


mhegazy:
@poke, sorry for the delay. Can you refresh the PR.

sheetalkamat:
@mhegazy  and @poke you shouldn't need this change any more as #257 handles this

PR #265, opened 2016-09-20
Use build script to generate grammars

sheetalkamat:
- TypeScript Grammar is written in yaml file
- TypeScriptReact Grammar has only additions and changes in its yaml file. The tmLanguage file for it is generated using build script
- Theme files are yaml files too. TypeScriptReact contains only additions to be applied to typescript theme


build/typings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "globalDependencies": {
+    "js-yaml": "registry:dt/js-yaml#3.5.2+20160316171810",
```
mhegazy: use `@types/js-yaml` and `@types/node` instead.

build/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "postinstall": "typings install && npm start",
+    "start": "tsc && node ./build.js"
+  },
+  "dependencies": {
+    "js-yaml": "^3.6.1",
+    "plist": "^2.0.1"
+  },
+  "devDependencies": {
+    "typings": "^1.3.3",
+    "typescript": "^1.6.2"
```
mhegazy: typescript@next

build/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "scripts": {
+    "postinstall": "typings install && npm start",
+    "start": "tsc && node ./build.js"
+  },
+  "dependencies": {
+    "js-yaml": "^3.6.1",
+    "plist": "^2.0.1"
+  },
+  "devDependencies": {
+    "typings": "^1.3.3",
```
mhegazy: use `@types` packages instead.

PR #362, opened 2016-11-17
Difference between function's Usage and Definition

pmkary:
When you're authoring a theme one thing with the new grammar really bothers and that's the fact that there is no difference between function's color when you're using a function and for when you're declaring a function:

screen shot 2016-11-18 at 1 26 53 am

I want my functions to be black but when I'm declaring them, they be red. I can because definition and usage have both exactly the same token names (`function.block.name.entity.meta.ts` to be exact)

screen shot 2016-11-18 at 1 27 57 am

screen shot 2016-11-18 at 1 28 19 am

My change is a very very simple one. I have added one more token name `definition` to the token for the function's definition so that one can choose only the definition or the usage and this change won't break any existing grammar as it's an incremental change to the previous version.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -641,7 +641,7 @@ repository:
       '2': { name: storage.modifier.async.ts }
       '3': { name: storage.type.function.ts }
       '4': { name: keyword.generator.asterisk.ts }
-      '5': { name: meta.definition.function.ts entity.name.function.ts }
+      '5': { name: meta.definition.function.ts entity.name.definition.function.ts }
     end: (?=;|\})|(?<=\})
     patterns:
     - include: '#comment'


sheetalkamat:
This is not needed. you can always look for ```meta.definition.method.ts entity.name.function.ts``` ```meta.definition.function.ts entity.name.function.ts``` to distinguish function definitions

pmkary:
@sheetalkamat ah thanks a lot for your help!

pmkary:
Okay that actually is not helping:

screen shot 2016-11-19 at 12 10 53 pm

Now just the opening `(`, `{` would get colored red.

sheetalkamat:
@pmkary  I am not sure what kind of theme changes you are making but only places we specifically set the scope name ``` meta.definition.method.ts entity.name.function.ts``` and ```meta.definition.function.ts entity.name.function.ts```` is for the function name. the ```(``` in there have scope name ```punctuation.definition.parameters.begin.ts``` 
Eg. look at https://github.com/Microsoft/TypeScript-TmLanguage/blob/master/tests/baselines/FunctionMethodOverloads.baseline.txt#L76

PR #367, opened 2016-11-22
Add scope meta.function-call for the function call

### LINKED ISSUE #364: Apply `function-call` scopes to all function calls
**[TS and JS Grammar Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version:**  0.4 / latest

**Code**

```ts
console.log
console.log()
process.exit()
fn()
```

In the above example, the `log` and `exit` are never matched on the `#function-call` directive within the tmLanguage. It also means `console.log` and `console.log()` have identical scoping on the `log`. 

The `#function-call` directive should add `function-call` scoping so that we can better highlight control flow.

The problem with `function.entity.name` is that it is hard to target with deep nesting scenarious. 

Adding `function-call` is a good step for regular function calls.
`name: entity.name.function.function-call.js`

I'd make a pull request but the fact is that scopes like `support.function.console` are hardcoded throughout, consuming the `#function-call` scoping.

I'm wondering if a re-ordering could fix this or if it's orthogonal to the direction of the this tmLanguage.

sheetalkamat:
Fixes #364

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -1378,7 +1378,7 @@ repository:
     - include: '#support-objects'
     - include: '#object-identifier'
     - include: '#punctuation-accessor'
-    - name: entity.name.function.ts
+    - name: variable.function.ts
       match: ([_$[:alpha:]][_$[:alnum:]]*)
     - include: '#comment'
     - name: meta.type.parameters.ts
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -19,7 +19,6 @@ repository:
     - include: '#string'
     - include: '#template'
     - include: '#comment'
-    - include: '#literal'
     - include: '#declaration'
     - include: '#switch-statement'
     - include: '#for-loop'
@@ -308,8 +307,7 @@ repository:
     - include: '#string'
     - name: entity.name.type.module.ts
       match: ([_$[:alpha:]][_$[:alnum:]]*)
-    - name: punctuation.accessor.ts
-      match: \.      
+    - include: '#punctuation-accessor'     
 
   import-equals-declaration: 
     patterns:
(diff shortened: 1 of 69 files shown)


sheetalkamat:
@nfour can you please see if this works for your ask in #364

nfour:
Thanks @sheetalkamat I will check this out soon!

nfour:
Managed to pull it down and take a look.

Some initial issues:

```js

Promise.all(streamEvents.map(async(event) => {
  // Get Event Details
}
```
In the above example, the `async` keyword and beyond, even the `comment` has `function-call` scopes.

```js
S._project.getVariablesObject(e.options.stage, e.options.region)
```

In the above, for some reason the `S` (any length of uppercase characters) gets `function-call`

sheetalkamat:
@nfour isn't that what you asked for.. The whole function call get the scope of function-call? Do you want the scope just for the function name as function-call? instead of whole function call expression?

nfour:
@sheetalkamat yeah exactly!

In `some.fn(() => { const a = 1; })` only `fn` should have `function-call`. This way you can style the control flow of function calling. The `function` scope would be applied to the body as it is now, though I've no preference for that as I don't see the point in styling function bodies.

sheetalkamat:
@nfour that is not the standard. Copying the section from https://www.sublimetext.com/docs/3/scope_naming.html 

> Function names, including the full path, and all parameters should receive the following scope. The name of the function or method should be variable.function, unless the function is scoped with support.function. 
◾ meta.function-call 

In addition to current change I am going to change the ```entity.name.function``` scope to ```variable.function``` in case of function call

sheetalkamat:
@aeschli I am changing the scope of function name in invocation from ```entity.name.function``` to ```variable.function``` as per the sublime scope naming convention link you had provided in the other bug. With this I think we should change the vs dark/dark+/light/light+ themes to color ```variable.function```  to yellow just like we use to color ```entity.name.function``` ?

capture

aeschli:
I haven't seen 'variable.function' in any grammar nor the built-in themes we have.
We can certainly adapt to it in our default themes, but I think most of the contributed themes just have a rule for 'variable' and will start rendering functions as variables.

nfour:
@sheetalkamat It doesn't really help the initial issue.

Consider the following in VSCODE.

in 

```js
const session = Session.decipher(event.headers[headerKey]);
```
- `Session` has:
  - `token meta js block variable other object function function-call`
- `decipher` has:
  - `token meta js block variable function function-call`
- `event` has:
  - `token meta js block variable support function dom function-call`

This doesn't actually add any value because there is still no way to style `decipher` differently to the other functions.

Because `function-call` etc. is inherited inside callbacks (parameters), then it's not actually a useful scope in the way that VSCode handles scopes at least.

In Atom sure, you'd be able to style this stuff trivially because it would have generated a nested hierarchy dom. VSCode can't implement the full tmTheme language feature set. Every scoped line is top level.

I think it's safe to say this tmLanguage targets VSCode.

aeschli:
@nfour VSCode will also switch to prober matching in the scope hierarchy in the next release.

nfour:
@aeschli well there goes my argument!

sheetalkamat:
I will close this pull request and associated bug and reopen if the fix is needed based on new scope matching

PR #372, opened 2016-11-30
Adds for-await-of

rbuckton:
Adds `await` keyword coloring for `for-await-of` to align with https://github.com/Microsoft/TypeScript/pull/12346.

### DIFF
--- a/TypeScript.tmLanguage
+++ b/TypeScript.tmLanguage
@@ -3161,7 +3161,7 @@
       <key>for-loop</key>
       <dict>
         <key>begin</key>
-        <string>(?&lt;!\.|\$)\b(for)\s*(\()</string>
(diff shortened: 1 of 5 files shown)


sheetalkamat:
👍

TypeScript.tmLanguage
```diff
@@ -3172,6 +3172,11 @@
           <key>2</key>
           <dict>
             <key>name</key>
+            <string>keyword.control.loop.ts</string>
```
sheetalkamat: I think this should be keyword.control.flow because that's how rest of the await is scoped.

rbuckton: I tried that but the coloring looked odd in the editor, plus `await` in this context isn't control flow, but modifies the loop behavior.

PR #442, opened 2017-04-05
Fixes #431 - spread operator on next line breaks highlighitng

### LINKED ISSUE #431: Spread operator can break JS highlighting.
_From @jeffmcmahan on March 12, 2017 19:5_

- VSCode Version: 1.10.2
- OS Version: 10.11.6

Steps to Reproduce:

1. Place this code in a JS file:

    ```js
    myFunc
    (...argName)
    
    function otherFunc() {} // <-- Not highlighted.
    ```
Remove the spread operator or the line break and the highlighting works again.

## Why fix this?
The parser used by VSCode is plainly in error. And FWIW, stuff like this is useful when working with higher-order functions:

```js

funcName
  (...someArgs)
  (someMore)
  (someMoreAgain)

```

_Copied from original issue: Microsoft/vscode#22483_

mjbvz:
Fixes #431

Moves `...` for arrow functions to the case where we require the `=>` to be on the same line, instead of allowing it to be on the next line


mjbvz:
Hey @sheetalkamat / @mhegazy, I don't see much recent activity on this repo so I wanted to check if I should loop in someone to review this PR

sheetalkamat:
Sorry it took so long to get back to this. I was on leave and got back just last week. 

This fix is not enough because the pattern like '(...' would never be recognized as arrow with this anymore. I will look into this and let you know how we can do better here. Also we need to add test case for this scenario.

sheetalkamat:
@mjbvz I think I got this to handle better with cc5499d and hence pushed that change out.

PR #449, opened 2017-05-04
Mark TSX Class Tags Differently from Normal Html Elements

### LINKED ISSUE #415: Mark JSX components differently from standard HTML tags
_From @tannerlinsley on January 18, 2017 19:14_

screen shot 2017-01-18 at 12 11 36 pm

This is no so in Atom (which is honestly the last thing keeping me from moving all the way to vscode):
screen shot 2017-01-18 at 12 13 50 pm

It doesn't seem to be fixed by using any theme, or installing the https://github.com/dzannotti/vscode-babel extension.

Any ideas?

_Copied from original issue: Microsoft/vscode#18752_

mjbvz:
Fixes #415

Adds an extra marker for tsx tags that are likly a javascript/typescript class instead of a standard html element. An element is marked as a class if it does not only contain lowercase letters

screen shot 2017-05-03 at 8 15 25 pm

### DIFF
--- a/TypeScriptReact.YAML-tmLanguage
+++ b/TypeScriptReact.YAML-tmLanguage
@@ -94,18 +94,18 @@ repository:
 
   jsx-tag-without-attributes:
     name: meta.tag.without-attributes.tsx
-    begin: (<)\s*(([a-z][a-z0-9]*|([_$a-zA-Z][-$\w.]*))(?<!\.|-))\s*(>)
-    end: (</)\s*(([a-z][a-z0-9]*|([_$a-zA-Z][-$\w.]*))(?<!\.|-))\s*(>)
+    begin: (<)\s*((?:[a-z][a-z0-9]*|([_$a-zA-Z][-$\w.]*))(?<!\.|-))\s*(>)
+    end: (</)\s*((?:[a-z][a-z0-9]*|([_$a-zA-Z][-$\w.]*))(?<!\.|-))\s*(>)
     beginCaptures:
       '1': {name: punctuation.definition.tag.begin.tsx}
       '2': {name: entity.name.tag.tsx}
-      '4': {name: support.class.component.tsx}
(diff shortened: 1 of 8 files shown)


TypeScriptReact.YAML-tmLanguage
```diff
@@ -94,16 +94,18 @@ repository:
 
   jsx-tag-without-attributes:
     name: meta.tag.without-attributes.tsx
-    begin: (<)\s*([_$a-zA-Z][-$\w.]*(?<!\.|-))\s*(>)
-    end: (</)\s*([_$a-zA-Z][-$\w.]*(?<!\.|-))\s*(>)
+    begin: (<)\s*(([a-z][a-z0-9]*|([_$a-zA-Z][-$\w.]*))(?<!\.|-))\s*(>)
+    end: (</)\s*(([a-z][a-z0-9]*|([_$a-zA-Z][-$\w.]*))(?<!\.|-))\s*(>)
```
sheetalkamat: Can you please change expressions in this PR from 
```
(([a-z][a-z0-9]*|([_$a-zA-Z][-$\w.]*))(?<!\.|-))
```

to 
```
((?:[a-z][a-z0-9]*|([_$a-zA-Z][-$\w.]*))(?<!\.|-))
```

This will help in easy reading when reading the capture groups.

tests/cases/Issue415.tsx
```diff
@@ -0,0 +1,7 @@
+(<View />);
```
sheetalkamat: Add below comment on the first line so that the baseline would be generated only against React grammar
```
// @onlyOwnGrammar - As this has jsx
```

TypeScriptReact.YAML-tmLanguage
mjbvz: Done

PR #504, opened 2017-08-18
Better support for template string grammar injections

### LINKED ISSUE #503: Better Support for injecting grammars into template strings
From: https://github.com/Microsoft/vscode/issues/5961

**Problem**
Many frameworks are using tagged template strings to embed other languages or construct domain specific languages. It is currently difficult to inject a grammar for these embedded languages into the the template string grammar

**Proposal**
Split the template string grammar element to better support injection. I'll submit a PR with this shortly

mjbvz:
Fixes #503

Allows for extension authors to more easily inject grammars for template strings. The change splits the template definition in two: a top level pattern to be targeted by the injection, and a standard implementation for js/ts template strings.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -1927,6 +1927,12 @@ repository:
 
   template:
     name: string.template.ts
+    begin: '(?=([_$[:alpha:]][_$[:alnum:]]*)?`)'
+    end: '(?<=`)'
+    patterns:
+     - include: '#standard-template-string'
+
+  standard-template-string:
     begin: '([_$[:alpha:]][_$[:alnum:]]*)?(`)'
     beginCaptures:
       '1': { name: entity.name.function.tagged-template.ts }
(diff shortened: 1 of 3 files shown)


mjbvz:
Here's the injection I originally tried to write:

```
{
	"fileTypes": [],
	"injectionSelector": "L:string.template.tsx",
	"patterns": [
		{
			"include": "#superjs-code-block"
		}
	],
	"repository": {
		"superjs-code-block": {
			"begin": "(?<=superjs`)",
			"end": "end",
			"patterns": [
				{
					"include": "#a"
				}
			]
		},
		"a": {
			"name": "a",
			"match": "a"
		}
	},
	"scopeName": "superjs.codeblock"
}
```

This should work for strings such as

~~~
superjs`a end`
~~~

But it does not because the `begin` rule with the negative lookahead puts the grammar into an endless loop.

mjbvz:
I've found a way to this without changing the grammar: https://github.com/mjbvz/vscode-js-template-string-grammar-injection-example

The solution was to introduce a noop rule that always consumes something inside the template string. I'll investigate to see if we can fix our engine to not require this

PR #511, opened 2017-09-06
Mark template expressions with meta.embedded.line scope

mjbvz:
**Problem**
VS Code themes currently hardcode a solution for reseting expression colorization inside of template string substitution expressions. This is not scalable since it js/ts specific

With https://github.com/Microsoft/vscode/issues/33120 I've proposed that VS Code themes reset styling inside of meta.embedded scopes to support this across multiple languages

**Fix**
Marks template substitution expressions with the additional `meta.embedded.block.ts` scope


TypeScript.YAML-tmLanguage
```diff
@@ -1952,6 +1952,7 @@ repository:
       '0': { name: punctuation.definition.template-expression.end.ts }
     patterns:
       - include: '#expression'
+    contentName: meta.embedded.block.ts
```
mjbvz: This should actually probably be `meta.embedded.line` now that I think about it more

PR #517, opened 2017-09-12
Fix jsdoc hang on default parameter value Fixes #515

### LINKED ISSUE #515: Hang on jsdoc comment with default value
**[TS and JS Grammar Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version:**  0.0.22 / latest

From https://github.com/Microsoft/vscode/issues/34164

**Code**

```ts
/**
 * @param {boolean} [settings.debug= - Print getUpdate results in console.
 */
```

**Bug**
Grammar hangs while parsing

mjbvz:
It seems that `[^*]` is causing a hang here. Replacing it with a more explicit set of allowed characters fixed things


TypeScript.YAML-tmLanguage
```diff
@@ -2297,7 +2297,7 @@ repository:
                 "(?:(?:\*(?!/))|(?:\\(?!"))|[^*\\])*?" |  # [foo="bar"] Double-quoted
```
sheetalkamat: Why not ^* in this line or below. We use the jsdoc grammar from https://github.com/atom/language-javascript/blob/master/grammars/jsdoc.cson#L283 Is it possible to get the fix their so that this one is not lost when trying to get the latest grammar from that repo. Thanks

mjbvz: Ok, let me try that. I don't fully understand what causes the hang thought. The freeze happens in native code so I haven't been able to debug into it.

Atom does not seem to have this problem but if we can get a good workaround, I can submit it upstream

mjbvz: @sheetalkamat How recently have you synced the jsdoc grammar? This issue is a lot like: https://github.com/atom/language-javascript/issues/495 

It seems to be a case of catastrophic backtracking instead of a pure hang. Adding text after the type comments causes more and more of a delay until eventually the window stops respinding

sheetalkamat: Updated now that there is fix included in their repo (about 4 days ago)

sheetalkamat:
Leave a comment

PR #519, opened 2017-09-13
Move generateScopes into `it` for tests

mjbvz:
**Bug**
In the tests, if one of the tests causes a hang, the entire test suite will hang. This is because `generateScopes` is being run in the `describe` block, instead of inside one of the tests themselves

**Fix**
Move `generateScopes` into the `it` block of the test

### DIFF
--- a/tests/test.ts
+++ b/tests/test.ts
@@ -22,27 +22,27 @@ ensureCleanGeneratedFolder();
 // Generate the new baselines
 for (const fileName of fs.readdirSync('cases')) {
     describe("Generating baseline for " + fileName, () => {
-        const text = fs.readFileSync(path.join('./cases', fileName), 'utf8');
-        const parsedFileName = path.parse(fileName);
-        const { markerScopes, wholeBaseline } = build.generateScopes(text, parsedFileName);
-        if (markerScopes) {
-            addTestCase(parsedFileName.name + '.txt', markerScopes);
-        }
-        addTestCase(parsedFileName.name + '.baseline.txt', wholeBaseline);
+        it('Comparing generated', () => {
+            const text = fs.readFileSync(path.join('./cases', fileName), 'utf8');
+            const parsedFileName = path.parse(fileName);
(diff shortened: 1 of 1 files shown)


tests/test.ts
```diff
@@ -22,27 +22,27 @@ ensureCleanGeneratedFolder();
...
-            addTestCase(parsedFileName.name + '.txt', markerScopes);
-        }
-        addTestCase(parsedFileName.name + '.baseline.txt', wholeBaseline);
+        it('Comparing generated', () => {
+            const text = fs.readFileSync(path.join('./cases', fileName), 'utf8');
+            const parsedFileName = path.parse(fileName);
+            const { markerScopes, wholeBaseline } = build.generateScopes(text, parsedFileName);
+            if (markerScopes) {
+                assertBaselinesMatch(parsedFileName.name + '.txt', markerScopes);
+            }
```
sheetalkamat: Problem with this is that it wont generate both baselines if the first one doesn't match..

mjbvz: Ok, I've move this back to two `it`s

tests/test.ts
```diff
@@ -24,25 +24,39 @@ for (const fileName of fs.readdirSync('cases')) {
...
         const text = fs.readFileSync(path.join('./cases', fileName), 'utf8');
         const parsedFileName = path.parse(fileName);
-        const { markerScopes, wholeBaseline } = build.generateScopes(text, parsedFileName);
-        if (markerScopes) {
-            addTestCase(parsedFileName.name + '.txt', markerScopes);
-        }
-        addTestCase(parsedFileName.name + '.baseline.txt', wholeBaseline);
+
+        const generateScopes = (() => {
+            let result: { markerScopes: string, wholeBaseline: string };
```
sheetalkamat: Nit.. May be add this to before() function in describe and hoist the markerScopes and wholeBaseline up.. then this function expression inside another one isn't needed.

PR #520, opened 2017-09-13
Use __dirname in tests so that cwd doesn't matter

mjbvz:
Update relative paths in tests to use `__dirname` when constructing the path. This allows running the tests without changing the cwd to `./tests`


sheetalkamat:
But tests are run using npm test from tests folder? How would anyone run tests from another folder?

mjbvz:
They can be run with mocha directly as well. I just configured vscode to run them and if you don't set the cwd, they fail

PR #533, opened 2017-10-13
Add new meta scope to mark attribute lists in tags

### LINKED ISSUE #532: Add new meta scope to identify jsx attribute list
From https://github.com/Microsoft/vscode/issues/36175

**Code**
For code such as 
```ts
<div
  className="foo"></div>
```

We'd like a new meta scope that specially marks the contents of the start tag itself (the parts between `<` and `>`). This is required to make sure we use the proper comment syntax in the editor

mjbvz:
Adds a new meta scope so that we can identity when we are in an attribute list for a tag

Fixes #532


TypeScriptReact.YAML-tmLanguage
```diff
@@ -115,6 +115,7 @@ repository:
         '2': { name: entity.name.tag.tsx }
         '3': { name: support.class.component.tsx }
       end: (?=[/]?>)
+      contentName: meta.tag.attributes.tsx
```
sheetalkamat: Are you sure that you also want to include #jsx-tag-attributes-illegal and #comment that gets applied inside this scope to have meta.tag.attributes.tsx ?

mjbvz: Yes, my intent was the have everything after the tag name but before the closing `>` scoped with  `meta.tag.attributes.tsx`

PR #545, opened 2017-11-30
fix error in TS.YAML in method-declaration-name

### LINKED ISSUE #544: Improper class method name highlighting
**[TS and JS Grammar Extension](https://atom.io/packages/atom-typescript) version:**  latest

**Code**

```ts
class foo {
  method() {
  }
}
```
In the above example 'method' is not highlighted

JNaftali:
fixes #544 

My first contribution, so apologies if I didn't update your tests or something. They were confusing - `cd test; npm test` had test failures and `npm run diff` threw a `/bin/sh: baselines: command not found` error so I'm not sure if I'm missing something. Still, the diff shows 3 lines changed - 1 in ts.yaml that I changed and the corresponding 2 lines in the generated ts.xml and tsx.xml files. So I'm fairly certain that things are as they should be.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -507,7 +507,7 @@ repository:
     patterns:
     - include: '#string'
     - include: '#array-literal'
-    - name: meta.definition.method.ts entity.name.function.ts
+    - name: meta.definition.method.entity.name.function.ts
       match: '{{identifier}}'
     - name: keyword.operator.optional.ts
       match: \?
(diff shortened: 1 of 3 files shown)


JNaftali:
I started suspecting as much as I looked around more. But that syntax in the cson for that Atom definitely results in the wrong css classes being applied. Sounds like a problem in the atom-typescript package, then?

sheetalkamat:
thanks correct. Closing this as working as intended..

damieng:
I can't see anything in the https://manual.macromates.com/en/language_grammars that would indicate names with spaces or multiple names are a valid syntax. Is there some other reference this project uses?

Issue #550, opened 2017-12-13
Typescript syntax highlighting fails for arrow functions with multiline arguments


mjbvz:
_From @christianbradley on December 12, 2017 18:57_

Visual Studio Code 1.18.1
Typescript 2.6.2

See example below:

```ts
export const foo = (firstVariable: string, secondVariable: number): string => {
    return "foo"
}

export const bar = (
    firstVariable: string,
    secondVariable: number,
    thirdVariable: string
): string => {
    return "baz"
}
```

_Copied from original issue: Microsoft/vscode#40129_

mjbvz:
Duplicate of #481 I believe

sheetalkamat:
Correct this is same as #481

sheetalkamat:
Released [VS Code Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version 0.0.33 that contains #552 which should address issues reported here. After letting it bake in that extension for 2-3 weeks, I could merge #552

EduardoRFS:
Same thing happen when using export default even with the extension

```typescript
// both arguments will lack color
export default (
  state: State = { rewards: [] },
  action: LoadAction | Action
) => {};

```typescript
// now it works
const defaultFunction = (
  state: State = { rewards: [] },
  action: LoadAction | Action
) => {};
```

Issue #539, opened 2017-11-08
No typescript highlighting for arrow function with line break before first param


mjbvz:
_From @arznaar on November 8, 2017 17:35_

Code below compiles correctly to javascript and runs without error.

- VSCode Version: 1.17.2
- Typescript version: 2.5.3
- OS Version: Windows 10 version 1703 build 15063.540
- Reproduces without extensions: Yes

Steps to Reproduce:

Write first param for the arrow function on the next line. Example: 
```
const goodMethod = (param1: string, param2: string) => {
}

const badMethod = (
    param1: string,
    param2: string
) => {
}
```

![vscode-bug-2017-11-08t17-15-00](https://user-images.githubusercontent.com/6916413/32563975-bdf93984-c4c3-11e7-8458-bef5c835b662.PNG)

_Copied from original issue: Microsoft/vscode#37837_

mjbvz:
This is a similar issue to #481 and some others. We currently do not handle multiline arrow function declarations well

arznaar:
>We currently do not handle multiline arrow function declarations well

It still can be reproduced in case of single arrow function
![vscode-bug-2017-11-08t21-29-00](https://user-images.githubusercontent.com/6916413/32567334-07e3b922-c4cc-11e7-9166-9dde2175a41a.PNG)

arznaar:
>We currently do not handle multiline arrow function declarations well

It still can be reproduced in case of single arrow function
![vscode-bug-2017-11-08t21-29-00](https://user-images.githubusercontent.com/6916413/32567334-07e3b922-c4cc-11e7-9166-9dde2175a41a.PNG)

arznaar:
>We currently do not handle multiline arrow function declarations well

It still can be reproduced in case of single arrow function
![vscode-bug-2017-11-08t21-29-00](https://user-images.githubusercontent.com/6916413/32567334-07e3b922-c4cc-11e7-9166-9dde2175a41a.PNG)

sheetalkamat:
This is duplicate of #481 and cant be fixed because of limitation of the tmLanguage syntax not being able to lookup beyond current line

arznaar:
It's so sad that highlighting works differently than compiler and we have such inconsistencies.

sheetalkamat:
[...]

Issue #524, opened 2017-09-22
standard color of function name is ignored when the arguments is on new line


mjbvz:
_From @isaklafleur on September 22, 2017 9:1_

- VSCode Version:
- OS Version:

Steps to Reproduce:

**I'm using prettier extension and it recommends to break on to separates lines. VSC don't color the variable name according to function color if the arguments are not on the same line.**

Paste this in editor and you will see the standard function variable color does appear on function `findAbbreviations`, but not on `replaceAbbreviationsWithExpansions` 

```js
const findAbbreviations = (string, arrayOfAbbreviations) => {
  const stringArray = stringSplit(string);
  const abbreviationMatches = [];
  arrayOfAbbreviations.forEach(item => {
    if (stringArray.indexOf(item.abbreviation) >= 0) {
      abbreviationMatches.push({
        abbreviation: item.abbreviation,
        expansion: item.expansion
      });
    }
  });
  return abbreviationMatches;
};

const replaceAbbreviationsWithExpansions = (
  string,
  abbreviations,
  findAbbreviations
) => {
  if (findAbbreviations.length > 0) {
    const abbreviationMatches = findAbbreviations(string, abbreviations);
    let newString = string;
    abbreviationMatches.forEach(item => {
      newString = newString.replace(item.abbreviation, item.expansion);
    });
    return newString;
  } else {
    return string;
  }
};
```

Reproduces without extensions: Yes/No
Yes

_Copied from original issue: Microsoft/vscode#34813_

mjbvz:
Simpler repo:

```js
const f1 = (x) => { }
const f2 = (
    x
) =>  {}
```

<img width="193" alt="screen shot 2017-09-22 at 2 07 30 pm" src="https://user-images.githubusercontent.com/12821956/30764364-671952b4-9f9f-11e7-85dd-0f49172e3741.png">

Scope of `f1` is:

```
entity.name.function.js
meta.definition.variable.js
meta.var-single-variable.expr.js
meta.var.expr.js
source.js
```


Scope of `f2` is:

```
variable.other.readwrite.js
meta.definition.variable.js
meta.var-single-variable.expr.js
meta.var.expr.js
source.js
```

sheetalkamat:
This is by design since on line 1 we cannot recognize the assignment to f2 as lambda expression

sheetalkamat:
[...]

Issue #507, opened 2017-08-24
Parameter types are not coloured if a newline is inserted before the first parameter


rcjsuen:
- VSCode Version: Code 1.15.0 (8b95971d8cccd3afd86b35d4a0e098c189294ff2, 2017-08-09T20:16:18.034Z)
- OS Version: Windows_NT x64 10.0.14393
---
Create this TypeScript file in VS Code.
```TypeScript
let ok: any = {
    func: (x: any, y: any): void => {
    }
}

let bug: any = {
    func: (
        x: any, y: any): void => {
    }
}
```
Note that the syntax highlighting is not quite right. You can even see in the above example that GitHub's Markdown syntax highlighting is also off.

![image](https://user-images.githubusercontent.com/15629116/29688037-4f7e161e-8959-11e7-9ac7-3bf729aa031e.png)

sheetalkamat:
This is by design and because of limitation of tmLanguage grammar syntax being not able to lookup multiple lines at once... So when looking at ( on the line of func we cant predict it being arrow and not just expression in parenthesis...

sheetalkamat:
Released [VS Code Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version 0.0.33 that contains #552 which should address issues reported here. After letting it bake in that extension for 2-3 weeks, I could merge #552

PR #552, opened 2017-12-14
Treat ( followed by new line as parameter context for arrow, falling back to expression

sheetalkamat:
Given that there are so many Issues that are getting duped to #481 and alike, i have created a branch that handles \( followed by line end as possibly arrow , this i am sure would lead to color few non arrow things as arrow (https://github.com/Microsoft/TypeScript-TmLanguage/pull/552/files#diff-a18148e1c0a6687d3019f4b439355bfbL290 will me marked as function instead of variable) but i am wondering if current behavior is better or the change here. 
Handles #481, #498, #507, #524, #539, #550, #558, #564, #574


arznaar:
Hm... It's better than before. Thank you!

But there are some cases, where it's still fails to highlight properly:
tshighlighterror

sheetalkamat:
Fixed and release [VS Code Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version 0.0.34 that contains the fix. Please give it a try.

arznaar:
Awesome! Thank you!

I have looked through my repo and found another place where highlighting can be improved:
tshighlighterror2

sheetalkamat:
Thank you for trying it out. I have fixed the reported issues and released [VS Code Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version 0.0.35 that contains the fix. Please give it a try.

jamesjryan:
@sheetalkamat works great via the limited use I gave it, what will it take to get this into VS Code upstream?

sheetalkamat:
@jamesjryan we need more feedback and need to watch out for the bugs to see if this is really viable fix. Even though I t fixes these scenarios, it would break multiple other scenarios and bugs reported, tweaks possible and cases breaking would be factor in merging this PR. Eg. today #579  was reported and I looked through our code and there are multiple instances where this just fails..

jamesjryan:
@sheetalkamat reporting a possible regression here - commenting a line using `cmd+/` within JSX now uses `// <SomeComponent />` as a comment, rather than `{/* <SomeComponent /> */}`.

sheetalkamat:
@jamesjryan can you please create a issue with sample code where this is failing. Thanks.

sheetalkamat:
FYI @mhegazy this is the PR that handles arrow function parameters that span on multiple lines

PR #557, opened 2018-01-15
support namespaced tsx

mtpc:
Add support for namespaced tsx as per jsx specification.
This is **not** a breaking change. All tests are green, I added an additional test for good measure.

The change is pretty simple, but it ends up being quite big due to the sheer amount of similar regexes:
A namespaced tag or attribute is defined by an identifier (namespace), followed by a colon, followed by another identifier (name), which gives us this addition: 
```coffeescript
///
(?: #start optional group
  ([_$a-zA-Z][-$\w.]*) #namespace
  (?<!\.|-) #prevent illegal end of namespace identifier
  : #colon
)? #end optional group
///
```


tests/baselines/pr557_namespacedJsx.baseline.txt
```diff
@@ -0,0 +1,444 @@
...
+                                                      source.tsx meta.var.expr.tsx meta.tag.tsx meta.tag.attributes.tsx
+                                                       ^^
+                                                       source.tsx meta.var.expr.tsx meta.tag.tsx punctuation.definition.tag.end.tsx
+                                                         ^
+                                                         source.tsx punctuation.terminator.statement.tsx
+>
+ ^
+ source.tsx
+>const valid2 = <namespaced:tag namespaced:boolean-attr>foo</namespaced:tag>;
+ ^^^^^
```
sheetalkamat: ```:tag``` and ```:boolean-attr``` is getting scoped here instead of just ```tag``` and ```boolean-attr```

tests/cases/pr557_namespacedJsx.tsx
```diff
@@ -0,0 +1,9 @@
+const valid1 = <standard-tag standard-attribute="foo" />;
```
sheetalkamat: Add comment ```// @onlyOwnGrammar - As this has jsx``` at the top to avoid scoping this for TypeScript without react support.

tests/baselines/pr557_namespacedJsx.baseline.txt
mtpc: It was intended, to highlight the colon the same color as the rest of the tag/attribute.
Would you prefer the colon to have its own scope? (`punctuation.separator.namespace.tsx`?)

sheetalkamat: yes please.

mtpc: This should be good

mtpc:
✓

PR #570, opened 2018-02-21
Fix issue with JSX attribute names not being colorized in VS

billti:
Fixes the issue discussed at https://developercommunity.visualstudio.com/content/problem/37434/jsx-attribute-no-longer-highlighted-in-dark-theme.html .

Note that tests were giving me massive failures on a fresh clone before any changes at all, so I'm just going to assume these are existing issues from out of date tests.

CC @mhegazy

### DIFF
--- a/TypeScript.YAML-tmTheme
+++ b/TypeScript.YAML-tmTheme
@@ -49,11 +49,11 @@ settings:
   settings: { vsclassificationtype: identifier }
 
 - scope: entity.name.tag
-  settings: { vsclassificationtype: xml literal - name }
+  settings: { vsclassificationtype: HTML Element Name }
 
 - scope: entity.other.attribute-name
-  settings: { vsclassificationtype: XML Attribute }
+  settings: { vsclassificationtype: HTML Attribute Name }
 
 - scope: meta.tag string.quoted, meta.tag string.quoted punctuation.definition.string, meta.tag string.quoted constant.character.escape
-  settings: { vsclassificationtype: xml literal - attribute value }
+  settings: { vsclassificationtype: HTML Attribute Value }
 ...
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,974 @@
+{
+  "name": "typescript-tmlanguage",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@types/chai": {
+      "version": "4.1.2",
+      "resolved": "https://registry.npmjs.org/@types/chai/-/chai-4.1.2.tgz",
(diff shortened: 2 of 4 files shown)


ChangedLater:
Hi, 
I am the original reporter in the link provided above. I have tested locally and these changes fix the issue above.
Thanks Bill

TypeScript.YAML-tmTheme
```diff
@@ -49,11 +49,11 @@ settings:
...
 - scope: entity.name.tag
-  settings: { vsclassificationtype: xml literal - name }
+  settings: { vsclassificationtype: HTML Element Name }
 
 - scope: entity.other.attribute-name
-  settings: { vsclassificationtype: XML Attribute }
+  settings: { vsclassificationtype: HTML Attribute Name }
 
 - scope: meta.tag string.quoted, meta.tag string.quoted punctuation.definition.string, meta.tag string.quoted constant.character.escape
-  settings: { vsclassificationtype: xml literal - attribute value }
```
sheetalkamat: You would also want to do this in [TypeScriptReact.Yaml-tmTheme](https://github.com/Microsoft/TypeScript-TmLanguage/blob/master/TypeScriptReact.YAML-tmTheme#L9)

package-lock.json
```diff
@@ -0,0 +1,974 @@
+{
```
sheetalkamat: Do not checking package-lock.json as its not related to this change.

TypeScript.YAML-tmTheme
billti: I wasn't sure what those scopes apply to, so I didn't want to risk making something worse as nobody is complaining about anything other than the attribute name.

package-lock.json
billti: It's a whitespace only change (line endings probably). There is no change that would impact anything.

TypeScript.YAML-tmTheme
sheetalkamat: I disagree. https://github.com/Microsoft/TypeScript-TmLanguage/blob/master/TypeScriptReact.YAML-tmTheme#L11 has attribute value which you changed in this PR to HTML Attribute Value, you want to do same on that line. I also think we should change the  { vsclassificationtype: xml literal - text } part as well since we changed everything else to HTML. this is just incomplete change.

package-lock.json
sheetalkamat: This file wasnt added to git repo earlier and is being added as part of your change.

billti: Then it was correct to add it (https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5).

TypeScript.YAML-tmTheme
dehghani-mehdi: Is there any way to change the color? (now JSX attrs have same color as the HTML tags and it's not good, tags are different from attributes)

PR #576, opened 2018-03-02
Fixes 575

mushishi78:
I do not have that much faith this is right. These tests are kinda hard to understand.


sheetalkamat:
This is not the right fix.

PR #583, opened 2018-03-16
Mark import paths with meta.import.path.ts

### LINKED ISSUE #582: Add a meta-scope to identify import paths
**Problem**
Similar problem to #546. In VS Code, quick suggestions are currently not enabled inside of `import` paths because we treat them as strings.

**Request**
Add a new meta scope to the string contents of import paths:

```
import 'foo'
        ---
         |               
  meta.import.path
```

This would allow us to specify that `meta.import.path` tokens should actually be treated like normal language syntax for quick suggestions and other basic language features

mjbvz:
Fixes #582

Adds a new meta scope to the contents of import paths: `meta.import.path`

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -936,6 +936,7 @@ repository:
     patterns:
     - include: '#import-path-string-double'
     - include: '#import-path-string-single'
+    - include: '#string'
 
   import-path-string-double:
     name: string.quoted.double.ts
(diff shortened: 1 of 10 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -932,6 +932,37 @@ repository:
     - name: variable.other.readwrite.alias.ts
       match: ({{identifier}})
 
+  import-path:
+    patterns:
+    - include: '#import-path-string-double'
+    - include: '#import-path-string-single'
```
mjbvz: Probably should fallback to `#string` here too

mjbvz:
Will fix by making vscode-textmate smarter: https://github.com/Microsoft/TypeScript-TmLanguage/issues/582#issuecomment-374390940

PR #588, opened 2018-03-23
Add s (single line) flag to js/ts regex

Titansmasher:
While not in the js/ts documentation, Chrome 64.0.3282.186 and Node 8.9.4 do support the `s` regex flag, as well as it being a suggested flag for [regex101](https://regex101.com). I feel that these are popular enough to justify adding it to the styling for the js/ts extensions.
I have tested IE, Edge and Firefox, however none of them appear to support the `s` flag.

I originally made this as a [PR for vscode](https://github.com/Microsoft/vscode/pull/46358) and was directed to this one

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -2030,10 +2030,10 @@ repository:
   regex:
     patterns:
     - name: string.regexp.ts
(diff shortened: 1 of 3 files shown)


Titansmasher:
In the test case for #346 (the one that is failing) if you replace the `s` character with a `g`,`i`,`m`,`u` or `y` then the same issue that was originally documented in #346 will occur. 
```javascript
for (let i = 0; i < b.length; i++) b[i] = Math.sin(TAU440 * t++^^/s * Math.sin(t++^^/s*.5))
```
to
```javascript
for (let i = 0; i < b.length; i++) b[i] = Math.sin(TAU440 * t++^^/s * Math.sin(t++^^/g*.5))
```
I dont think this introduces a new issue, rather accidentally highlighted an existing one. I dont feel I have the knowledge to be able to fix that underlying issue, should I open up a bug report for that and submit the `s` flag as a suggestion instead?

sheetalkamat:
Closing this in favor of 8610134

PR #617, opened 2018-05-25
Separate control flow from loops and try catch

karl:
I've been working on my own theme and I'd really like to be able to target `break`, `continue`, and `throw` separately from the rest of the `keyword.control.trycatch` and `keyword.control.loop` scopes.

I'd like to be able to highlight the `break`, `continue` and `throw` keywords separately as they all allow you to break out of the current flow of execution.

I'm not entirely sure of the best way to go about doing this, so I thought I'd open a speculative PR and see if I could get some feedback.

I've moved `break` and `continue` from the `keyword.control.loop` scope into `keyword.control.loop.flow`. I'm hoping this change avoids breaking existing themes while still giving us the ability to target these keywords separately.

For `throw` I've moved it from `keyword.control.trycatch` to `keyword.control.trycatch.flow`.

If I've gone completely off track here please let me know! I'm totally happy if these PR gets closed and there needs to be more discussion.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -153,10 +153,14 @@ repository:
     patterns:
     - include: '#switch-statement'
     - include: '#for-loop'
+    - name: keyword.control.trycatch.flow.ts
+      match: '{{startOfIdentifier}}(throw){{endOfIdentifier}}'
     - name: keyword.control.trycatch.ts
-      match: '{{startOfIdentifier}}(catch|finally|throw|try){{endOfIdentifier}}'
+      match: '{{startOfIdentifier}}(catch|finally|try){{endOfIdentifier}}'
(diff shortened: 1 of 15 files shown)


sheetalkamat:
There haven't been any asks for this. Usually the existing level of granularity is sufficient and most of the grammars do only till that scope level. 

Given that there arent many grammars that break it down to this level of granularity, your theme will not work for many other grammars out there, example atom javascript grammar. We wouldnt be taking this change until we are convinced that it is used more widely by other grammars as well.

karl:
@sheetalkamat Thanks for the response! I understand that there isn't much interest in this yet, so no worries about closing the PR.

I've created an issue to document my interest in this and try and explain why it would be useful. We'll see if it gathers interest from anyone else. Happy to have it closed if it doesn't garner much interest!

Thanks again for the quick response.

PR #620, opened 2018-05-31
Add lib reference syntax

rbuckton:


rbuckton:
I'm working on that, but I'm having some trouble running tests locally due to an issue with node compile-time bindings for Oniguruma.

rbuckton:
@sheetalkamat I've added a test for this. Any other comments?

PR #655, opened 2018-08-21
Allow underscore in jsx attribute names

### LINKED ISSUE #648: JSX properties red if there is an underscore in prop name
_From @marsiancba on August 17, 2018 13:50_

This is probably some bug in TSX syntax highlighting:

image

```
<Test
    a=''
    b_c='111'
    d=''
/>

function Test(props: { a: string, b_c: string, d: string }) { return <span /> };
```

_Copied from original issue: Microsoft/vscode#56685_

mjbvz:
Fixes #648


tests/cases/Issues648.tsx
```diff
@@ -0,0 +1,5 @@
+<Test
```
sheetalkamat: Add line at the beginning of file to avoid generating ts grammar baseline
```typescript
// @onlyOwnGrammar - As this has jsx
```

sheetalkamat:
Also add test case for #649

PR #682, opened 2018-11-08
Adding azure pipelines

damccorm:
[This pipeline](https://dev.azure.com/TypeScript-TmLanguage/TypeScript-TmLanguage/_build?definitionId=2) points at my fork, you can see its successfully building.

**Main Advantages:**
1. Azure Pipelines lets us build and test on Windows, Linux, and macOS (one CI service instead of multiple)
2. Azure Pipelines provides unlimited minutes and 10 parallel jobs for open source projects (which means faster builds)
3. Great integration with GitHub

**Next steps once this PR is submitted**
1. Point pipeline at source repo instead of my fork
2. Transfer ownership of pipeline to azuredatastudio maintainers
3. I’ll create another PR to add a status badge
4. Once we see its working, remove travis/appveyor since they'll be redundant

### DIFF
--- a/azure-pipelines-template.yml
+++ b/azure-pipelines-template.yml
@@ -7,19 +7,13 @@ steps:
 - script: npm install
   displayName: 'npm install'
 
-- script: npm install -g mocha
-  displayName: 'Install mocha'
-
-- script: npm install -g mocha-junit-reporter
-  displayName: 'Install mocha junit reporter'
-
 - script: npm run build
   displayName: 'Build'
 
-- script: mocha tests/test.js --reporter mocha-junit-reporter
+- script: npm run test
   displayName: 'Test'
 
 - task: PublishTestResults@2
   inputs:
-    testResultsFiles: '**/test-results.xml'    
+    testResultsFiles: '**/xunit.xml'    
   condition: succeededOrFailed()
--- a/azure-pipelines-template.yml
+++ b/azure-pipelines-template.yml
@@ -1,7 +1,7 @@
 steps:
 - task: NodeTool@0
   inputs:
-    versionSpec: '8.9'
+    versionSpec: '$(node_version)'
   displayName: 'Install Node.js'
 
 - script: npm install
--- a/azure-pipelines-template.yml
+++ b/azure-pipelines-template.yml
@@ -1,0 +1,25 @@
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '8.9'
+  displayName: 'Install Node.js'
+
+- script: npm install
+  displayName: 'npm install'
+
+- script: npm install -g mocha
+  displayName: 'Install mocha'
+
+- script: npm install -g mocha-junit-reporter
+  displayName: 'Install mocha junit reporter'
+
+- script: npm run build
+  displayName: 'Build'
+
+- script: mocha tests/test.js --reporter mocha-junit-reporter
+  displayName: 'Test'
+
+- task: PublishTestResults@2
+  inputs:
+    testResultsFiles: '**/test-results.xml'    
+  condition: succeededOrFailed()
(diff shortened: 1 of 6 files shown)


sheetalkamat:
Added @weswigham since he handles pipelines. Thanks.

azure-pipelines-template.yml
```diff
@@ -0,0 +1,25 @@
...
+  displayName: 'Install Node.js'
+
+- script: npm install
+  displayName: 'npm install'
+
+- script: npm install -g mocha
+  displayName: 'Install mocha'
+
+- script: npm install -g mocha-junit-reporter
+  displayName: 'Install mocha junit reporter'
```
weswigham: Why the whole custom mocha install/deep test file run instead of a simple `npm test`?

weswigham: Ah, because of the xml result collection.

weswigham: I'd consider editing https://github.com/Microsoft/TypeScript-TmLanguage/blob/master/package.json#L12 instead of doing lots of tool-specific work - it's not like the xml file is going to fail any other CI providers. On that note, using _just_ the junit reporter is kind eck - it ruins the logfile output for humans (and removes all progress indication). Setting up a [multi-reporter](https://www.npmjs.com/package/mocha-multi-reporters) structure would be much better.

damccorm: Good call - I've updated appropriately with a multi-reporter structure within the npm test call

azure-pipelines-template.yml
```diff
@@ -0,0 +1,19 @@
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '8.9'
```
weswigham: Can this be templated so we get a node `8` and a latest stable node build, like the Travis yaml specifies? We don't need to test every version on every platform; but checking both versions in addition to each platform would be nice (I'd check each platform on latest stable and 'nix only for node `8`).

On a related note: are there any fast build options we can opt in to? Like Travis had a container opt in (sudo: false); from what I understand the new non-preview 'nix images aren't containerized builds anymore. Mostly just curious here.

damccorm: @weswigham I updated so that it gives the latest version of 11, unfortunately I don't think the node tool installer can check for the latest major version though (it will automatically bump with minor versions now though). Also, it looks like Windows builds are failing on Node 11, npm install seems to fail, it works fine on node 8 though.

damccorm: https://dev.azure.com/TypeScript-TmLanguage/TypeScript-TmLanguage/_build/results?buildId=28&view=logs

weswigham: Can the vm be provisioned with python 2 instead of python 3?

damccorm: @weswigham when you get a chance would you mind taking a look at this again? I think it should be good to go.

weswigham:
@sheetalkamat this looks fine to me - if you're OK with it, we can use it.

damccorm:
@sheetalkamat @weswigham thanks for getting this in! I just created the new Pipeline pointing at this repo (instead of my fork) [here](https://dev.azure.com/ms/TypeScript-TmLanguage/_build?definitionId=15) and added you both to the owning team. In a moment I'll also create a PR to add a build status badge to the README. At this point there should just be 3 remaining steps:

1. Add Pipelines as a PR check
2. (optionally) accept my build status badge PR
3. (optionally) Remove Travis after you see Pipelines working for a bit

sheetalkamat:
@damccorm The build is failing on azure with some error. Can you please take a look at whats going on. Thanks

damccorm:
Sorry I missed this. Taking a look now, it seems like the builds are running again? We had an LSI that I think could have been related to this, so maybe that was the cause?

weswigham:
We swapped out the service connection the build was using, since the build was failing because the one it was using seemed to suddenly lack any permissions.

Issue #693, opened 2018-12-05
Highlight @example jsdoc body as source code


mjbvz:
**[TS and JS Grammar Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version:**  0.0.47 / latest

**Code**
For a js doc body such as:

```ts
/**
 * @example
 * 1 + 1
 */
const a = 1
```

We mark the body of the `@example` block specially but do not actually syntax highlight it. I added an injection grammar in VS Code that does this: https://github.com/Microsoft/vscode/blob/master/extensions/typescript-basics/syntaxes/ExampleJsDoc.injection.tmLanguage.json

May be worth including in the core grammar. Let me know if you agree. The injection grammar will not work as is, we're relying on another injection grammar that handles separating the leading stars of the doc block from the documentation content itself

mjbvz:
This is for highlighting the content `1 + 1` as typescript. Right now, it is just treated as a single token

sheetalkamat:
@mjbvz I don't think that's as easy as it seems .. I think that need to be handled by editors (by removing starting `*` on each line). Without that the results will be too unpredictable.

mjbvz:
We use a while rule to do this, see: https://github.com/Microsoft/vscode/blob/master/extensions/typescript-basics/syntaxes/ExampleJsDoc.injection.tmLanguage.json

It consumes the leading `*` on each line so that the regular js/ts grammar runs without modification inside the comment

sheetalkamat:
I still don't understand how multiline example is going to work. Where do you consume leading `*` wasn't able to figure that out.

mjbvz:
Looks like the logic that was moved into this file: https://github.com/Microsoft/vscode/blob/master/extensions/typescript-basics/syntaxes/MarkdownDocumentationInjection.tmLanguage.json

The `while` rule is run on every line until it fails. In this case, it consumes the leading `*` so that rules inside the `while` never see it.

PR #740, opened 2019-04-08
Consume leading docblock *s

### LINKED ISSUE #467: Multiline JSDoc Types
From https://github.com/Microsoft/vscode/issues/27777

**[TS and JS Grammar Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version:**  0.0.20

**Code**

```js
/**
 * @param {{
 *  z: number
 * }} x
 */
function foo(x){}
```

Current highlighting: 

screen shot 2017-05-31 at 2 49 02 pm

There seems to be some debate about whether this is supported or not: https://github.com/jscs-dev/node-jscs/issues/245 but it seems closure supports this syntax

mjbvz:
Fixes #467

- Adds a begin/while rule so that doc blocks also consume the leading *. This approach has been used by VS Code to support highlighting example blocks (see #693)

- Remove the `invalid.illegal.type.jsdoc` rule since we now can highlight multiline types better

Before:

Screen Shot 2019-04-08 at 2 24 39 PM

After:

Screen Shot 2019-04-08 at 2 24 54 PM

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -2671,7 +2671,11 @@ repository:
       endCaptures:
         '0': { name: punctuation.definition.comment.ts }
       patterns:
-      - include: '#docblock'
+      - contentName: comment.block.documentation.body.ts
+        begin: (?<=/\*\*)([^*]|\*(?!/))*$
+        while: (^|\G)\s*\*(?!/)(?=([^*]|[*](?!/))*$)
+        patterns:
+        - include: '#docblock'
     - name: comment.block.ts
       begin: (/\*)(?:\s*((@)internal)(?=\s|(\*/)))?
       beginCaptures:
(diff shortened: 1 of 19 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -2671,7 +2671,11 @@ repository:
       endCaptures:
         '0': { name: punctuation.definition.comment.ts }
       patterns:
-      - include: '#docblock'
+      - contentName: comment.block.documentation.body.ts
+        begin: (?<=/\*\*)([^*]|\*(?!/))*$
+        while: (^|\G)\s*\*(?!/)(?=([^*]|[*](?!/))*$)
```
sheetalkamat: The while pattern does not work in sublime and results in error and hence cannot be applied in this repo as it sublime typescript package also uses this grammar.

mjbvz: Unfortunately is no way to implement this without begin/while. I can keep this grammar modification VS Code but it means that other consumers of this grammar (like VS) will not benefit

PR #764, opened 2019-08-01
Note that jsdoc grammar is unmaintained upstream

SamB:
They're using tree-sitter now, and don't want tmLanguage patches anymore.

(See https://github.com/atom/language-javascript/pull/645#issuecomment-496795093.)


orta:
Agree, thanks

Issue #522, opened 2017-09-21
Support option chaining syntax highlighting


mjbvz:
**[TS and JS Grammar Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version:**  0.0.22 / latest

From https://github.com/Microsoft/vscode/issues/34776#event-1259771216

**Feature request** 
Support highlighting the proposed [option chaining](https://github.com/tc39/proposal-optional-chaining) syntax for JS/TS

```ts
const a = {
  b: {
    c: {
      d: {
        e: 1,
        g: [1, 2, 3],
      },
    },
    f: null,
  },
};

const brokenSyntaxHighlighting = () => {
  const array = a?.b?.c?.d?.g;

  /**
   * No TypeEror, but syntax highlighting is broken
   */
  const test = a?.b?.f?.f;

  for (let i = 0; i < array.length; i++) {
    console.log(array[i]);
  }
};

brokenSyntaxHighlighting();
```

![30707451-31af05c4-9f04-11e7-9b4f-12e6ff3ad2f8](https://user-images.githubusercontent.com/12821956/30709245-72ca5908-9eb6-11e7-8a41-5ecba2d79816.png)

TS support for this is tracked by: https://github.com/Microsoft/TypeScript/issues/16

sheetalkamat:
This feature is still in stage 1. Will look at it once it gets into stage 3

zalishchuk:
@sheetalkamat Is it really hard to implement syntax support for this feature? We have to use something like [idx](https://github.com/facebookincubator/idx) for these purposes which is not always convenient.

oychao:
I need this very hardly.

zalishchuk:
@sheetalkamat can you please update https://github.com/sheetalkamat/TypeScript-TmLanguage-VsCode Visual Code extension?

fungilation:
Time to revisit this, with React Native 56 turning integrating optional chaining support, now in babel 7.

This is the best thing about coffeescript and so much more convenient than the verbose alternative. But broken syntax highlighting in VSCode is getting in the way of full adoption.

mehulmpt:
Well currently you can sort of get around this by disabling VSCode's internal javascript linter and using something like eslint, but obviously you lose some features of vscode linter like dimming of unused variables/imports, etc.

fungilation:
I see in my VSCode default config:
```
"eslint.enable": true,
```

[...]

PR #768, opened 2019-08-16
add nullish coalescing support

Kingwl:
a part of #522 

Could i test in my local vscode?

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -218,6 +218,7 @@ repository:
     - include: '#arrow-function'
     - include: '#paren-expression-possibly-arrow'
     - include: '#cast'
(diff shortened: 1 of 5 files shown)


Kingwl:
https://github.com/microsoft/TypeScript/pull/32883

TypeScript.YAML-tmLanguage
```diff
@@ -218,6 +218,7 @@ repository:
     - include: '#arrow-function'
     - include: '#paren-expression-possibly-arrow'
     - include: '#cast'
+    - include: '#nullish-coalesce'
```
sheetalkamat: Instead of this you want to just add the matching rule to `expression-operators` and change ternary to see lookup next to `?` is not `?`

PR #779, opened 2019-10-11
Use 'contentName' instead of 'name' to mark embedded jsx expressions

mjbvz:
Jsx expressions currently use `name` to mark the entire expresion (including the opening and closing braces) as `meta.embedded.expression`. This causes VS Code to incorrectly treat the entire expression as JavaScript for basic language commands such as commenting.

This fix instead changes jsx expressions to only mark the expression itself (and not the braces) as `meta.embedded.expression`. This lets VS Code correctly handle commenting in a few edge cases


mjbvz:
From https://github.com/microsoft/vscode/issues/82224

PR #798, opened 2020-01-02
Improve scopes of type cast declarations

forivall:
I'm working on some syntax highlighting to better disambiguate what is type annotations and what is actual executable code, and I found that scope information in type casts were missing.

Let me know if there's better scope namings for any of these, I based it on what names made sense based on tm scope standards & what's elsewhere in this tmLanguage definition. Or, why this change can't be made (no worries!).


tests/baselines/ArrowFunctionInsideTypeAssertion.baseline.txt
```diff
@@ -67,37 +67,37 @@ Grammar: TypeScript.tmLanguage
                    ^
                    source.ts meta.var.expr.ts meta.objectliteral.ts meta.object.member.ts cast.expr.ts meta.brace.angle.ts
                     ^
-                    source.ts meta.var.expr.ts meta.objectliteral.ts meta.object.member.ts cast.expr.ts meta.type.function.ts meta.parameters.ts punctuation.definition.parameters.begin.ts
+                    source.ts meta.var.expr.ts meta.objectliteral.ts meta.object.member.ts cast.expr.ts meta.type.cast.ts meta.type.function.ts meta.parameters.ts punctuation.definition.parameters.begin.ts
```
sheetalkamat: As you can see there is already`cast.expr.ts` so this change isn't needed.

forivall: `cast.expr.ts` includes the surrounding `<` and `>`, and i have `meta.type.cast.ts` include only the contents. (see diffs at https://github.com/microsoft/TypeScript-TmLanguage/pull/798/files#diff-d5222b162266b45fd998419db416fdddR1577 and https://github.com/microsoft/TypeScript-TmLanguage/pull/798/files#diff-d5222b162266b45fd998419db416fdddR1591

In my use case, i want to change the syntax highlighting of the contents of the cast, but not the surrounding punctuation (`<` `>`) or the `as`.

That was something i forgot to mention: since `cast.expr.ts` is used for `<MyType>variable` style-casts, should `variable as MyType` style casts also use `cast.expr.ts`, or is `meta.type.cast.ts` preferable? If they are switched to `cast.expr.ts`, should the `as` also have `cast.expr.ts`?

If i switch `variable as MyType` casts from `meta.type.cast.ts`, should the `as` also be scoped as `cast.expr.ts`, since

forivall:
Something i forgot to mention: since `cast.expr.ts` is used for `<MyType>variable` style-casts, should `variable as MyType` style casts also use `cast.expr.ts`, or is `meta.type.cast.ts` preferable? If they are switched to `cast.expr.ts`, should the `as` also have `cast.expr.ts`?

tests/baselines/ArrowFunctionInsideTypeAssertion.baseline.txt
sheetalkamat: The scenario you are looking for can be achieved by omitting `cast.expr.ts meta.brace.angle.ts`. We want to keep this simple and not add unnecessary scopes. Most themes do not care about the difference and hence we wont be accepting this change

forivall:
I removed it that part. Personally, i'm not currently satisfied, since scope exclusion is limited with the currently available scope selectors so verbose workarounds are required (for example: https://github.com/microsoft/vscode-textmate/issues/52#issuecomment-519567440 ). But hopefully that'll eventually be fixed.

tests/baselines/Issue314.baseline.txt
```diff
@@ -36,9 +36,9 @@ Grammar: TypeScript.tmLanguage
                  ^
                  source.ts meta.var.expr.ts
                   ^
-                  source.ts meta.var.expr.ts entity.name.type.ts
+                  source.ts meta.var.expr.ts meta.type.cast.ts entity.name.type.ts
```
sheetalkamat: Please use `cast.expr.ts` as scope instead of `meta.type.cast.ts`

TypeScript.YAML-tmLanguage
```diff
@@ -1375,7 +1375,7 @@ repository:
       match: '{{startOfIdentifier}}(as)\s+(const)(?=\s*([,}]|$))'
       captures:
         '1': { name: keyword.control.as.ts }
-        '2': { name: storage.modifier.ts }
+        '2': { name: storage.modifier.const.type.ts }
```
sheetalkamat: please do not change scope for modifier `const`. Instead add `name` as `cast.expr.ts`

TypeScript.YAML-tmLanguage
```diff
@@ -1561,7 +1561,7 @@ repository:
       match: \s*(<)\s*(const)\s*(>)
       captures:
         '1':  { name: meta.brace.angle.ts }
-        '2':  { name: storage.modifier.ts }
+        '2':  { name: storage.modifier.const.type.ts }
```
sheetalkamat: Same here

TypeScript.YAML-tmLanguage
```diff
@@ -1624,11 +1624,12 @@ repository:
     - match: '{{startOfIdentifier}}(as)\s+(const)(?=\s*($|[;,:})\]]))'
       captures:
         '1': { name: keyword.control.as.ts }
-        '2': { name: storage.modifier.ts }
+        '2': { name: storage.modifier.const.type.ts }
```
sheetalkamat: Same here

TypeScript.YAML-tmLanguage
```diff
@@ -1624,11 +1624,12 @@ repository:
...
     - begin: '{{startOfIdentifier}}(as)\s+'
       beginCaptures:
         '1': { name: keyword.control.as.ts }
       end: (?=$|^|[;,:})\]]|\|\||\&\&|({{startOfIdentifier}}(as)\s+)|(\s+\<))
+      contentName: meta.type.cast.ts
```
sheetalkamat: Name this as `cast.expr.ts` and use `name` instead of `contentName` since cast expression starts at `as`

TypeScript.YAML-tmLanguage
```diff
@@ -1375,7 +1375,7 @@ repository:
       match: '{{startOfIdentifier}}(as)\s+(const)(?=\s*([,}]|$))'
       captures:
         '1': { name: keyword.control.as.ts }
-        '2': { name: storage.modifier.ts }
+        '2': { name: storage.modifier.const.type.ts }
     - name: meta.object.member.ts
       begin: '{{startOfIdentifier}}(as)\s+'
       beginCaptures:
```
sheetalkamat: Add name for this as `cast.expr.ts` as well

PR #861, opened 2021-03-09
Add 'abstract' keyword to constructor type syntax.

rbuckton:
Added in https://github.com/microsoft/TypeScript/pull/36392

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -2443,17 +2443,19 @@ repository:
   type-fn-type-parameters:
     patterns:
     # constructor type
-    - begin: '{{startOfIdentifier}}(new)\b(?=\s*\<)'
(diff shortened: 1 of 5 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -2443,9 +2443,10 @@ repository:
   type-fn-type-parameters:
     patterns:
     # constructor type
-    - begin: '{{startOfIdentifier}}(new)\b(?=\s*\<)'
+    - begin: '{{startOfIdentifier}}(?:(abstract)\s+)?(new)\b(?=\s*\<)'
```
sheetalkamat, line 2446: Need to add it at https://github.com/microsoft/TypeScript-TmLanguage/pull/861/files#diff-f920686662480359e6be0b6446e4dde96319b28e5135f8216b43c389023fc157L2454

sheetalkamat (changes requested):

sheetalkamat (approved):

PR #862, opened 2021-03-09
Add 'd' flag from the RegExp Match Indices proposal

rbuckton:
Adds `d` flag to Regular Expressions, per https://github.com/tc39/proposal-regexp-match-indices per [22.2.3.2.2 RegExpInitialize ( obj, pattern, flags )](https://arai-a.github.io/ecma262-compare/?pr=1713&id=sec-regexpinitialize), Step 5:

> 
image


sheetalkamat (changes requested):
Missing baseline update

sheetalkamat (approved):

PR #899, opened 2022-02-17
adding 'auto' to .gitattributes

noshlag:
This should resolve the issue with the ThirdPartyNotices.txt file showing up as "modified" because of line ending changes.

See the documentation here for the change that occurs with the 'auto' value for text.
https://git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-Settostringvalueauto


noshlag:
I thought these changes addressed the issue, but after some local testing, I'm no longer certain.

PR #902, opened 2022-03-18
Updated to support 'infer..extends'

rbuckton:
This adds support for `infer..extends` (https://github.com/microsoft/TypeScript/pull/48112) to our .tmLanguage.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -2332,8 +2332,8 @@ repository:
     - include: '#type-parameters'
     - include: '#type-tuple'
     - include: '#type-object'
-    - include: '#type-conditional'
     - include: '#type-operators'
+    - include: '#type-conditional'
     - include: '#type-fn-type-parameters'
     - include: '#type-paren-or-function-parameters'
     - include: '#type-function-return-type'
@@ -2519,6 +2519,7 @@ repository:
   type-operators:
     patterns:
     - include: '#typeof-operator'
+    - include: '#type-infer'
     # Handle the object types followed by | or & operator as { can be used as end of Type in many places and this avoids tripping that
     - begin: ([&|])(?=\s*\{)
       beginCaptures:
@@ -2536,10 +2537,17 @@ repository:
       match: '{{startOfIdentifier}}keyof{{endOfIdentifier}}'
     - name: keyword.operator.ternary.ts
       match: (\?|\:)
-    - name: keyword.operator.expression.infer.ts
-      match: '{{startOfIdentifier}}infer(?=\s+[_$[:alpha:]])'
     - name: keyword.operator.expression.import.ts
       match: '{{startOfIdentifier}}import(?=\s*\()'
+
+  type-infer:
+    patterns:
+    - match: '{{startOfIdentifier}}(infer)\s+({{identifier}}){{endOfIdentifier}}(?:\s+(extends){{endOfIdentifier}})?'
+      name: meta.type.infer.ts
+      captures:
+        '1': { name: keyword.operator.expression.infer.ts }
+        '2': { name: entity.name.type.ts }
+        '3': { name: keyword.operator.expression.extends.ts }
 
   type-predicate-operator:
     patterns:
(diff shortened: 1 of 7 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -2536,11 +2537,30 @@ repository:
...
+    patterns:
+    # try to match infer..extends on a single line
+    - match: '{{startOfIdentifier}}(infer)\s+({{identifier}})\s+(extends){{endOfIdentifier}}'
+      name: meta.type.infer.ts
+      captures:
+        '1': { name: keyword.operator.expression.infer.ts }
+        '2': { name: entity.name.type.ts }
+        '3': { name: keyword.operator.expression.extends.ts }
+    # alternatively, match across multiple lines
+    - begin: '{{startOfIdentifier}}infer(?=\s+[_$[:alpha:]])'
```
sheetalkamat, line 2553: Can you add test that matches this?

package.json
```diff
@@ -23,6 +23,7 @@
     "mocha": "latest",
     "plist": "latest",
     "typescript": "latest",
+    "vscode-grammar-updater": "^1.0.4",
```
sheetalkamat, line 26: Why do we need this..

tests/cases/inferTypes.ts
```diff
@@ -1,4 +1,5 @@
 type ReturnType<T extends Function> = T extends ((...args: any[]) => infer R) | (new (...args: any[]) => infer R) ? R : any;
 type ArgumentType<T extends (x: any) => any> = T extends (a: infer A) => any ? A : any;
 type X1<T extends { x: any, y: any }> = T extends { x: infer X, y: infer Y } ? [X, Y] : any;
-type B1<S> = S extends A1<infer T, infer U> ? [T, U] : never;
```
sheetalkamat, line 4: Test baselines have some failures

sheetalkamat (changes requested):

package.json
rbuckton: Ah, meant to revert that. I was trying to simplify testing the tmLanguage in VS Code since it looks like they switched to pre-generating a .JSON file from the tmLanguage for TypeScript so I couldn't just copy/paste.

TypeScript.YAML-tmLanguage
rbuckton: Added

TypeScript.YAML-tmLanguage
```diff
@@ -2536,11 +2537,30 @@ repository:
...
+      beginCaptures:
+        '0': { name: keyword.operator.expression.infer.ts }
+      end: '(?=[;),}\]:?\-\+\>''"`|&]|\|\||\&\&|\!\=\=)'
```
sheetalkamat, line 2556: Why do we need to match this across multiple lines. is it common if not i think we should just not add it..  
Concern here is the ending regex not containing line end or beginning or line in some form... On Single line matching it ends at extends so why not same? so why do we need to match more than extends.. 
We try to end things in type space with new line to ensure we aren't coloring things as `type` because of many ambiguity and textmate's requirement to look ahead only line at a time..

rbuckton: Primarily because there's no NLT restriction between `infer T` and `extends`. I can drop the multi-line support if it doesn't seem worthwhile.

sheetalkamat: i think its better to skip it..  We can always add it if many people are interested and we see it as more value

TypeScript.YAML-tmLanguage
```diff
@@ -2536,11 +2537,18 @@ repository:
...
     - name: keyword.operator.ternary.ts
       match: (\?|\:)
-    - name: keyword.operator.expression.infer.ts
-      match: '{{startOfIdentifier}}infer(?=\s+[_$[:alpha:]])'
     - name: keyword.operator.expression.import.ts
       match: '{{startOfIdentifier}}import(?=\s*\()'
 
+  type-infer:
+    patterns:
+    - match: '{{startOfIdentifier}}(infer)\s+({{identifier}})\s+(extends){{endOfIdentifier}}'
```
sheetalkamat, line 2545:
```suggestion
    - match: '{{startOfIdentifier}}(infer)\s+({{identifier}})\s+(?:(extends){{endOfIdentifier}})?'
```

I think you want this to make sure infer T is recognized as well

rbuckton: I wonder if it would be easier to treat `extends` as just another type operator, and remove it from `type-conditional` as well.

sheetalkamat (approved):

PR #958, opened 2023-01-31
[fix] fix #956: the case like `as const satisfies T`

### LINKED ISSUE #956: `as const satisfies` idiom is not highlighted correctly
Hello, team!

I'm looking into this grammar definition and found `as cont satisfies` idiom is not highlighted correctly. See the example bellow.

**[TS and JS Grammar Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.typescript-javascript-grammar) version:**  0.0.55 / latest

**Code**

```ts
const a = [1, 2, 3] satisfies ReadonlyArray<number>; // Array<number>
const b = [1, 2, 3] as ReadonlyArray<number>; // ReadonlyArray<number>
const c = [1, 2, 3] as const satisfies ReadonlyArray<number>; // readonly [1, 2, 3]
```

Screenshot:

Screenshot 2023-01-30 at 14 19 22

I'm expecting the `satisfies` keyword in `as const satisfies` to be handled in the same color as `as` because it's the same part of grammar.

gfx:
It seems that there's a typo in the language definition file, which causes #956. 

fix #956

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -1240,7 +1240,7 @@ repository:
       captures:
         '1': { name: keyword.control.type.ts }
         '2': { name: variable.other.readwrite.alias.ts }
-      
+
 
   #control statements and loops
   switch-statement:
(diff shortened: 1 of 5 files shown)


gfx:
@microsoft-github-policy-service agree

sheetalkamat (approved):

PR #959, opened 2023-02-03
Updates tmThemes to match latest VSCode colors

armanio123:
Updates typescript and typescriptreact themes to match VSCode colorization.

There are still slight differences between them but this is very close to match all colors.

This is how it looks now on Visual Studio:
Before --> After
image

### DIFF
--- a/TypeScript.YAML-tmTheme
+++ b/TypeScript.YAML-tmTheme
@@ -46,10 +46,11 @@ settings:
 - scope: entity.name.type.enum.ts
   settings: { vsclassificationtype: enum name }
 
-- scope: entity.name.type, meta.template.expression.ts, entity.other.inherited-class.ts
+- scope: meta.template.expression.ts, entity.other.inherited-class.ts
   settings: { vsclassificationtype: identifier }
+- scope: entity
 - scope: variable, meta.definition.variable.name, support.variable, entity.name.variable, constant.other.placeholder
-  settings: { vsclassificationtype: parameter name }
+  settings: { vsclassificationtype: local name }
 - scope: entity.name.function, support.function, support.constant.handlebars, source.powershell variable.other.member, entity.name.operator.custom-literal
   settings: { vsclassificationtype: method name }
 
--- a/TypeScript.YAML-tmTheme
+++ b/TypeScript.YAML-tmTheme
@@ -48,7 +48,6 @@ settings:
 
 - scope: meta.template.expression.ts, entity.other.inherited-class.ts
   settings: { vsclassificationtype: identifier }
-- scope: entity
 - scope: variable, meta.definition.variable.name, support.variable, entity.name.variable, constant.other.placeholder
   settings: { vsclassificationtype: local name }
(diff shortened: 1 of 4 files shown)


zkat:
@armanio123 What are the remaining differences!

TypeScript.YAML-tmTheme
```diff
@@ -42,8 +46,12 @@ settings:
 - scope: entity.name.type.enum.ts
   settings: { vsclassificationtype: enum name }
 
-- scope: entity.name.function, entity.name.type, meta.template.expression.ts, variable, entity.other.inherited-class.ts
+- scope: entity.name.type, meta.template.expression.ts, entity.other.inherited-class.ts
```
sheetalkamat, line 49:
```suggestion
- scope: meta.template.expression.ts, entity.other.inherited-class.ts
```
given entity.name.type is classified as type above .

TypeScript.YAML-tmTheme
```diff
@@ -42,8 +46,12 @@ settings:
 - scope: entity.name.type.enum.ts
   settings: { vsclassificationtype: enum name }
 
-- scope: entity.name.function, entity.name.type, meta.template.expression.ts, variable, entity.other.inherited-class.ts
+- scope: entity.name.type, meta.template.expression.ts, entity.other.inherited-class.ts
   settings: { vsclassificationtype: identifier }
+- scope: variable, meta.definition.variable.name, support.variable, entity.name.variable, constant.other.placeholder
+  settings: { vsclassificationtype: parameter name }
```
sheetalkamat, line 52: how come `const x = 10` x is classified as parameter name ?

sheetalkamat (changes requested):

TypeScript.YAML-tmTheme
armanio123: Fixed typo. This should be `local name`;

TypeScript.YAML-tmTheme
armanio123: Thanks, good catch. Fixed.

armanio123:
> @armanio123 What are the remaining differences?

This are some of the differences I have identified:
- Punctuation, is much more colorful in VSCode. 
- Tags punctuation in JSX, they are darker on VSCode.
- The `delete` keyword. I think this is an actual issue on VSCode tho. I decided to fix it for this template.
- Function arrow. I have not found how VSCode colorize this, I continue to search for it and send another PR when found.
- The `foo` function on this screen is misleading. In reality it will show as `type`. I'm submitting a different PR to match it.
- The `enum` members are slightly different color. There's no matching color on the VS side. Nontheless, I decided to still setup the token to allow users to change the color through configuration.

This is the comparison between them:
VSCode --> VS after this PR.
image

armanio123:
There are some other differences when comparing with the default light colors, but overall I still think is an improvement.
- Variables. 
- Numbers.
- Escape character.

Before PR --> This PR --> VS Code
image

sheetalkamat (approved):

PR #963, opened 2023-02-16
Fixes jsx tags colorization and improves JSDoc comments

armanio123:
Fixed Jsx tag colorzation. They display inconsistent colors for open and closed tags. 

Improved JSDoc comments:
- Variables now display `local name` color instead of `identifier`.
- JSDoc directives now use `HTML Element Name`.

Before --> After
image


zkat:
this looks so much better tbh

sheetalkamat (approved):

PR #982, opened 2023-05-11
Fix `undefined` keyword highlighting

armanio123:
Removes `constant.language.undefined.ts` from the scope so that the `constant.language` scope can pick it up.

Fixes: https://developercommunity.visualstudio.com/t/JavaScript:-undefined-not-highlighted-/10358005

VS before :
image

VS After:
image

### DIFF
--- a/TypeScript.YAML-tmTheme
+++ b/TypeScript.YAML-tmTheme
@@ -51,7 +51,7 @@ settings:
 - scope: entity.name.function, support.function, support.constant.handlebars, source.powershell variable.other.member, entity.name.operator.custom-literal
   settings: { vsclassificationtype: method name }
 
-- scope: constant.language.undefined.ts, variable.language.arguments.ts, support.type.object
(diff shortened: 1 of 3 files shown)


sheetalkamat (changes requested):
undefinded is constant so you might instead update the theme file instead. 
I am not sure if this change will break others including vscode.

sheetalkamat (changes requested):
undefinded is constant so you might instead update the theme file instead. 
I am not sure if this change will break others including vscode.

armanio123:
By removing the scope, I'm actually making the `constant.language` scope, identify and pick up the syntax highlight for `undefined`. In Visual Studio all constants are mapped to `keyword` classification. There's no classification defined for "constants" yet.

In regard to the change, I think it will not break others. As far as I know, this file is only used by Visual Studio.

sheetalkamat (approved):
Sorry .. for some reason I read this as yaml for grammar instead of theme . This is correct change

PR #1018, opened 2024-02-13
Make `regex` lookahead check more strict

RedCMD:
Fixes https://github.com/microsoft/vscode/issues/204813
> Regex on new line breaks js/ts highlighter

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -140,7 +140,7 @@ variables:
     {{functionLikeType}} |
     (:\s*(=>|{{matchingParenthesis}}|(<[^<>]*>)|[^<>(),=])+={{functionOrArrowLookup}})
   arrowFunctionEnd: (?==>|\{|(^\s*(export|function|class|interface|let|var|{{usingKeyword}}|{{awaitUsingKeyword}}|const|import|enum|namespace|module|type|abstract|declare)\s+))
-  regexpTail: ([dgimsuy]+|(?![\/\*])|(?=\/\*))(?!\s*[a-zA-Z0-9_$])
+  regexpTail: (?!\*)[dgimsuy]*((?!\/)|(?=\/\*))(?!\s*[a-zA-Z0-9_$])
   completeRegexp: \/(?![\/*])(?=(?:[^\/\\\[]|\\.|\[([^\]\\]|\\.)*\])+\/{{regexpTail}})
 
 patterns:


mjbvz (approved):

sheetalkamat (changes requested):
Please add testcase which this fixes

jimmy-zhening-luo:
@RedCMD , bump. Here's an example of valid TypeScript that transpiles to correct JavaScript but fails highlighting in my IDE:

# Example

```typescript
protected url(string: string) {
  const regex = /^(([^:/?#]+):)?(\/\/([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?/u;
  
  // Hereafter, highlighting broken
  const parts = regex.exec(string);
  
  if (parts === null)
    throw new Error("Input unparseable as URL");
  else
    return {
      scheme: parts[2],
      host: parts[4],
      path: parts[5],
      query: parts[7],
      fragment: parts[9],
    };
}
```

# Environment

- TypeScript versions 5.4.5 to 5.6.2 checked
- Targets ES6
- Strictest compiler options
- Is correctly highlighted in VSCode, but fails in my IDE which depends on the latest TypeScript TmLanguage

RedCMD:
@jimmy-zhening-luo what IDE do you use?
can you send a screenshot?
should prob make a new bug report

RedCMD:
Before:
image

After:
image

TypeScript.YAML-tmLanguage
```diff
@@ -140,7 +140,7 @@ variables:
     {{functionLikeType}} |
     (:\s*(=>|{{matchingParenthesis}}|(<[^<>]*>)|[^<>(),=])+={{functionOrArrowLookup}})
   arrowFunctionEnd: (?==>|\{|(^\s*(export|function|class|interface|let|var|{{usingKeyword}}|{{awaitUsingKeyword}}|const|import|enum|namespace|module|type|abstract|declare)\s+))
-  regexpTail: ([dgimsuy]+|(?![\/\*])|(?=\/\*))(?!\s*[a-zA-Z0-9_$])
+  regexpTail: (?!\*)[dgimsuy]*((?!\/)|(?=\/\*))(?!\s*[a-zA-Z0-9_$])
```
jakebailey, line 143: Are you sure this shouldn't also be applied to other tmLanguage files as done in #1032?

The PR also seems to be missing updates to the tmLanguage files themselves.

jakebailey (changes requested):

TypeScript.YAML-tmLanguage
RedCMD: is there no auto github workflow?

jakebailey: I don't think so, you should just be able to run `npm run build:grammar`.

It'd be good if CI verified that this had been run.

Mazahuacar (approved):

PR #1030, opened 2024-05-24
Support "Arbitrary Module Namespace Identifiers"

evanw:
This PR implements [Arbitrary Module Namespace Identifiers](https://github.com/tc39/ecma262/pull/2154), which is a JavaScript syntax feature that was added to JavaScript over two years ago (in ES2022). It allows import and export aliases to be strings instead of identifiers.

**Before this PR:**

Notice that the string content is incorrectly colored. In particular, the characters `default` in the middle of the string are confusingly highlighted as a keyword. The preceding `type` keyword is also not highlighted as a keyword when it should be.

**After this PR:**

All strings are highlighted as strings, and the `type` keyword is highlighted as a keyword.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -15,7 +15,6 @@ variables:
   propertyIdentifier: '\#?{{identifier}}'
   constantPropertyIdentifier: '\#?{{constantIdentifier}}'
   quotedStrings: (\'([^\'\\]|\\.)*\')|(\"([^\"\\]|\\.)*\")|(\`([^\`\\]|\\.)*\`)
-  aliasString: \'(?:[^\'\\]|\\.)*\'|\"(?:[^\"\\]|\\.)*\"
   nonIdentifierPropertyName: '{{quotedStrings}}|(\[([^\[\]]|\[[^\[\]]*\])+\])'
   label: ({{identifier}})\s*(:)
   hexNumber: \b(?<!\$)0(?:x|X)[0-9a-fA-F][0-9a-fA-F_]*(n)?\b(?!\$)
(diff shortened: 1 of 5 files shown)


evanw:
Update: This syntax feature has now been [added to TypeScript](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#support-for-arbitrary-module-identifiers) and [shipped in VSCode](https://code.visualstudio.com/updates/v1_94#_typescript-56). However, VSCode shipped it with broken syntax highlighting because this PR wasn't merged. Merging this PR would fix syntax highlighting for this feature.

TypeScript.YAML-tmLanguage
```diff
@@ -15,6 +15,7 @@ variables:
   propertyIdentifier: '\#?{{identifier}}'
   constantPropertyIdentifier: '\#?{{constantIdentifier}}'
   quotedStrings: (\'([^\'\\]|\\.)*\')|(\"([^\"\\]|\\.)*\")|(\`([^\`\\]|\\.)*\`)
+  aliasString: \'(?:[^\'\\]|\\.)*\'|\"(?:[^\"\\]|\\.)*\"
```
sheetalkamat, line 18: You dont need this. just use quotedStrings instead. Grammar here can be more permissive than actual and thats ok.

tvassociation31:
{
  "id": "chatcmpl-12345",
  "object": "chat.completion",
  "created": 1677652288,
  "model": "gpt-4",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "Hi"
      },
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 10,
    "completion_tokens": 2,
    "total_tokens": 12
  }
}

PR #1054, opened 2025-05-19
Align the specification of `typeof` operator within expression operators

### LINKED ISSUE #978: `typeof` with `<` breaks syntax highlighting
**[JavaScript and TypeScript Nightly](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-next) version:** v5.1.20230416

**Code**

```ts
typeof x < ''; function f() {}
```

For context, here's the whole code where I noticed it:
```js
var gt = typeof globalThis < "u" ? globalThis : typeof window < "u" ? window : typeof global < "u" ? global : typeof self < "u" ? self : {};
```

Actual behaviour: `typeof` with `<` or `<=` gets interpreted as the begin of a type parameter and breaks all syntax highlighting after it.
image
image

Expected behaviour: should be a relational operator.
image

### LINKED ISSUE #1050: Incorrect Javascript syntax highlighting with typeof value and less than (or equal to) operator
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.96.2
- OS Version: macOS 15.3.2 (24D81), Darwin arm64 24.3.0

Steps to Reproduce:

```
x = typeof y < 'z';
if (x) {
} else {
}
```

Image

Image

zyoshoka:
Fixes #978
Fixes #1050

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -1744,7 +1744,8 @@ repository:
       match: '{{startOfIdentifier}}instanceof{{endOfIdentifier}}'
     - name: keyword.operator.new.ts
       match: '{{startOfIdentifier}}new{{endOfIdentifier}}'
-    - include: '#typeof-operator'
+    - name: keyword.operator.typeof.ts
+      match: '{{startOfIdentifier}}typeof{{endOfIdentifier}}'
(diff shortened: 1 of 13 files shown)


zedd3v:
@mjbvz

PR #1073, opened 2026-05-18
Fix const type parameters in TSX

### LINKED ISSUE #979: TypeScript's const modifier on type parameteres breaks syntax highlighting in `.tsx` files

phryneas:
fixes #979

I tried to reuse existing tests as far as possible - this should finally fix synax highlighing for `<const Foo extends Bar>` kind type parameters.


tests/cases/constTypeParameter.ts
```diff
@@ -1,4 +1,3 @@
-// @onlyOwnGrammar - this has type parameters that would be jsx in tsx files
```
phryneas, line 1: This test now works with both ts and tsx

tests/cases/constTypeParameter.ts
```diff
@@ -10,7 +9,6 @@ class C1<const T> {
     foo<const U>(x: U) { return x; }
 }
 const C2 = class <const T> {}
-const fx1 = <const T>(x: T) => x;
```
phryneas, line 13: This line is ambiguous in tsx - moved into a separate test

tests/baselines/constTypeParameter.baseline.txt
```diff
@@ -476,51 +469,6 @@ Grammar: TypeScript.tmLanguage
                             source.ts meta.var.expr.ts meta.class.ts punctuation.definition.block.ts
                              ^
                              source.ts meta.var.expr.ts meta.class.ts punctuation.definition.block.ts
->const fx1 = <const T>(x: T) => x;
```
phryneas, line 479: This just moved to the other file.

